PR #5, opened 2024-10-08
[feat] refactor remain ratio to TDMSampler and remove root node in positive sampled data

tiankongdeguiji:


chengaofei (approved):

tzrec/datasets/sampler.py
```diff
@@ -745,49 +764,73 @@ def get(self, input_data: Dict[str, pa.Array]) -> Dict[str, pa.Array]:
...
         pos_ids = np.concatenate((pos_non_leaf_ids, ids), axis=1)
-
         pos_fea_result = self._parse_nodes(pos_nodes)[1:]
-        pos_result_dict = dict(zip(self._attr_names[1:], pos_fea_result))
+
+        # randomly select layers to keep
+        if self._remain_ratio < 1.0:
+            remain_layer = np.random.choice(
+                range(1, self._max_level - 1),
+                int(self._remain_ratio * (self._max_level - 1)),
```
jjbbong: The expression should multiply by (self._max_level - 2) because the root node must be droped out.

tzrec/datasets/sampler.py
```diff
@@ -745,49 +764,73 @@ def get(self, input_data: Dict[str, pa.Array]) -> Dict[str, pa.Array]:
...
+                replace=False,
+                p=self._remain_p,
+            )
+        else:
+            remain_layer = np.array(range(1, self._max_level - 1))
+        remain_layer.sort()
+
+        if self._remain_ratio < 1.0:
+            pos_fea_index = np.concatenate(
+                [remain_layer + j * (self._max_level - 2) for j in range(batch_size)]
```
jjbbong: The expression should be `remain_layer - 1 + j * (self._max_level - 2) for j in range(batch_size)` because there's no edge to root

tzrec/datasets/sampler.py
```diff
@@ -745,49 +764,73 @@ def get(self, input_data: Dict[str, pa.Array]) -> Dict[str, pa.Array]:
...
-        batch_size = len(ids)
-        cum_layer_num = np.cumsum([0] + self._layer_num_sample[:-1])
-        same_user_index = np.concatenate(
+        cum_layer_num = np.cumsum(
+            [0]
+            + [
+                self._layer_num_sample[i] if i in remain_layer else 0
+                for i in range(self._max_level - 1)
+            ]
+        )
```
jjbbong, line 819: `cum_layer_num`should not be changed，still set `cum_layer_num = np.cumsum([0] + self._layer_num_sample[:-1])`, because the negative sampling operator still samples from all layers, and the offset should take all layers into account.

tzrec/datasets/sampler.py
jjbbong, line 1: The parameter `expand_factor` of `self._pos_sampler` should be `self._max_level - 2` because the edges connected to the root node were directly removed during edge construction.

tzrec/datasets/sampler.py
```diff
@@ -745,49 +764,73 @@ def get(self, input_data: Dict[str, pa.Array]) -> Dict[str, pa.Array]:
...
+            pos_fea_result = [
+                pos_fea_result[i].take(pos_fea_index) for i in range(num_fea)
+            ]
 
         # negative sample layer by layer.
         neg_fea_layer = []
-        for i in range(1, self._max_level):
-            neg_nodes = self._neg_sampler_list[i].get(pos_ids[:, i], pos_ids[:, i])
+        for i in np.append(remain_layer, self._max_level - 1):
+            neg_nodes = self._neg_sampler_list[i - 1].get(pos_ids[:, i], pos_ids[:, i])
```
jjbbong: `neg_nodes = self._neg_sampler_list[i - 1].get(pos_ids[:, i-1], pos_ids[:, i-1])` because pos_ids doesn't have root node

tzrec/datasets/sampler.py
tiankongdeguiji: negative sampling operator only samples remain layers

jjbbong: Oh, my mistake

chengaofei (approved):

PR #24, opened 2024-11-04
[feat] make docker compat with gpu driver 470

tiankongdeguiji:


chengaofei (approved):

docker/Dockerfile
yjjinjie, line 1: 为什么要卸载？ 没太明白 470可以兼容？那515这些驱动不用这几个包了吗
另外sed -i '/^Requires-Dist: nvidia-/d' 这个是在干嘛？

tiankongdeguiji: We should use the CUDA library located in /usr/local/cuda instead of the NVIDIA toolkit installed through PyPI for compatibility with the 470 driver. 

We use the command sed -i '/^Requires-Dist: nvidia-/d' to remove the CUDA PyPI dependency for the Torch wheel.

You can refer to the official NVIDIA documentation on CUDA compatibility at: https://docs.nvidia.com/deploy/cuda-compatibility/.

PR #57, opened 2024-12-09
[feat]rank model supports sample weight

eric-gecheng:
Adapt rank model loss calculation to support sample weight, especially JRC loss.

### DIFF
--- a/tzrec/loss/jrc_loss.py
+++ b/tzrec/loss/jrc_loss.py
@@ -102,9 +102,10 @@ class JRCLoss(_Loss):
             logits_neg + ((1 - neg_session_mask) + (1 - diag_neg) * y_neg) * -1e9
         )
         loss_neg = self._ce_loss(logits_neg, neg_diag_label)
-        loss_pos = loss_pos * pos_num / batch_size
-        loss_neg = loss_neg * neg_num / batch_size
+
         if self._reduction != "none":
+            loss_pos = loss_pos * pos_num / batch_size
+            loss_neg = loss_neg * neg_num / batch_size
             ge_loss = loss_pos + loss_neg
         else:
             ge_loss = torch.zeros_like(labels, dtype=torch.float)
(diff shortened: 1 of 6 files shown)


tzrec/models/rank_model.py
```diff
@@ -183,6 +185,9 @@ def _loss_impl(
     ) -> Dict[str, torch.Tensor]:
```
tiankongdeguiji, line 188: The sample_weight_name parameter, similar to label_name, needs to be an input for the _loss_impl function. In a multi-task ranking model, it's essential to specify distinct sample weights for each loss function.

tiankongdeguiji (changes requested):

tzrec/loss/jrc_loss.py
```diff
@@ -88,8 +102,15 @@ def forward(self, logits: Tensor, labels: Tensor, session_ids: Tensor) -> Tensor
             logits_neg + ((1 - neg_session_mask) + (1 - diag_neg) * y_neg) * -1e9
         )
         loss_neg = self._ce_loss(logits_neg, neg_diag_label)
+        loss_pos = loss_pos * pos_num / batch_size
+        loss_neg = loss_neg * neg_num / batch_size
+        if self._reduction != "none":
+            ge_loss = loss_pos + loss_neg
+        else:
+            ge_loss = torch.zeros_like(labels, dtype=torch.float)
```
tiankongdeguiji, line 111: add a test for  reduction=='mean' in jrc_loss_test.py

tiankongdeguiji (changes requested):

tzrec/models/rank_model.py
eric-gecheng: fixed

tzrec/loss/jrc_loss.py
eric-gecheng: added a test for reduction == 'none' since 'mean' is the default option.

tiankongdeguiji (approved):
LGTM

PR #59, opened 2024-12-10
py310& py311 : use lower system version compiled torch-trt(larec) for system compatibility(train&predict)

yjjinjie:


tiankongdeguiji (approved):
LGTM

PR #65, opened 2024-12-16
[bugfix]fix calculation of weighted loss

eric-gecheng:


tzrec/models/match_model.py
```diff
@@ -233,7 +233,9 @@ def _loss_impl(
             label = _zero_int_label(pred)
         losses[loss_name] = self._loss_modules[loss_name](pred, label)
         if self._sample_weight:
-            losses[loss_name] = torch.mean(losses[loss_name] * sample_weight)
+            losses[loss_name] = torch.sum(
```
tiankongdeguiji: use torch.mean(xxx)/torch.mean(xxx) is better, loss and sample weight could be too large

tiankongdeguiji (changes requested):

tiankongdeguiji (approved):
LGTM

PR #83, opened 2025-01-07
[feat] add dual augmented two-tower model

eric-gecheng:


tzrec/models/match_model.py
```diff
@@ -234,10 +235,39 @@ def _loss_impl(
         else:
             label = _zero_int_label(pred)
         losses[loss_name] = self._loss_modules[loss_name](pred, label)
+
+        if all(
```
tiankongdeguiji: move 238-265 to dat.py is better.
```python
def loss(
        self, predictions: Dict[str, torch.Tensor], batch: Batch
    ) -> Dict[str, torch.Tensor]:
    losses = super().loss(predictions, batch)
    # do DAT augment losses
    return losses
```

tzrec/models/match_model.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, Alibaba Group;
+# Copyright (c) 2024-2025, Alibaba Group;
```
tiankongdeguiji, line 1: add unit test for DATModel in dat_test.py

docs/source/models/recall.rst
```diff
@@ -6,3 +6,4 @@
 
    dssm
    tdm
+   dat
```
tiankongdeguiji, line 9: add DAT to README.md

tiankongdeguiji (changes requested):

tzrec/models/match_model.py
eric-gecheng: fixed

tzrec/models/match_model.py
eric-gecheng: added

docs/source/models/recall.rst
eric-gecheng: added

tiankongdeguiji (approved):
LGTM

PR #110, opened 2025-02-20
[bugfix] fix create_fg_json invalid option force_update_resource

chengaofei:
fix_create_fg_json_invalid_option_force_update_resource

### DIFF
--- a/tzrec/tools/create_fg_json.py
+++ b/tzrec/tools/create_fg_json.py
@@ -57,8 +57,8 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--force_update_resource",
-        action="store_false",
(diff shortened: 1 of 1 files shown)


tzrec/tools/create_fg_json.py
```diff
@@ -57,13 +57,13 @@
     )
     parser.add_argument(
         "--force_update_resource",
-        type=bool,
+        action="store_false",
         default=True,
```
tiankongdeguiji: action="store_true",default=False, is better

tiankongdeguiji (changes requested):

tiankongdeguiji (approved):
LGTM

PR #157, opened 2025-04-09
[feat] optimize mind model

eric-gecheng:
Notes: 
1. optimize creation and scaling for the routing_logit tensor
2. optimize the iteration of dynamic routing, capturing gradient after iteration 
3. adjust MindUserTower's MLP modules, the inner layers and output layers are extracted separately
4. Add bias hyper-parameter for the MLP module. For sequence feature,  bias is not used

### DIFF
--- a/tzrec/models/mind.py
+++ b/tzrec/models/mind.py
@@ -142,12 +142,15 @@ class MINDUserTower(MatchTower):
             hist_seq_feas = grp_hist_seq
 
         high_capsules = self._capsule_layer(hist_seq_feas, grp_hist_len)
+        high_capsules_mask = norm(high_capsules, dim=-1) != 0.0
 
         # concatenate user feature and high_capsules
         user_feature = torch.unsqueeze(user_feature, dim=1)
         user_feature_tile = torch.tile(user_feature, [1, high_capsules.shape[1], 1])
         user_interests = torch.cat([user_feature_tile, high_capsules], dim=-1)
+        user_interests = user_interests * high_capsules_mask.unsqueeze(-1).float()
         user_interests = self._concat_mlp(user_interests)
+        user_interests = user_interests * high_capsules_mask.unsqueeze(-1).float()
 
         if self._output_dim > 0:
             user_interests = self.output(user_interests)
--- a/tzrec/models/mind.py
+++ b/tzrec/models/mind.py
@@ -72,7 +72,9 @@ class MINDUserTower(MatchTower):
                 hidden_units=tower_config.user_mlp.hidden_units[0:-1],
                 activation=tower_config.user_mlp.activation,
                 use_bn=tower_config.user_mlp.use_bn,
-                dropout_ratio=tower_config.user_mlp.dropout_ratio[0],
+                dropout_ratio=tower_config.user_mlp.dropout_ratio[0]
+                if tower_config.user_mlp.dropout_ratio
+                else None,
             )
             self.user_mlp_out = nn.Linear(
                 self.user_mlp.output_dim(), tower_config.user_mlp.hidden_units[-1]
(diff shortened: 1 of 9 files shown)


tzrec/modules/capsule.py
```diff
@@ -112,14 +114,10 @@ def dynamic_routing(
...
         """
-        routing_logits = torch.concat(
-            [
-                torch.randn_like(inputs, dtype=torch.float32),
-                torch.randn_like(inputs, dtype=torch.float32),
-            ],
-            dim=-1,
-        )
-        routing_logits = routing_logits[:, :, : self._max_k].to(inputs.device)
+        routing_logits_tmp = torch.zeros_like(inputs)[
```
tiankongdeguiji: 1. Why change randn to rand ?
2. Why origin implementation pass fx trace ut ?
3. May be
```python
@torch.fx.wrap
def _rand_k_channel_like(x: torch.Tensor, k: int) -> torch.Tensor:
    return torch.rand(x.size()[:-1] + torch.Size([k]), device=x.device, dtype=x.dtype)

routing_logits = _rand_k_channel_like(inputs, self._max_k) 
```
is better.

eric-gecheng: various fixes have been applied in below commits.

tzrec/models/mind.py
```diff
@@ -104,23 +154,38 @@ def forward(self, batch: Batch) -> torch.Tensor:
         grp_hist_seq = user_feature_dict[self._hist_group_name + ".sequence"]
         grp_hist_len = user_feature_dict[self._hist_group_name + ".sequence_length"]
 
-        user_feature = self.user_mlp(user_feature_dict[self._group_name])
+        if self.user_mlp_out:
```
tiankongdeguiji, line 158: why we need self.user_mlp_out? self.output do not have activation, should we need another layer without activation?

tzrec/models/mind.py
```diff
@@ -263,8 +338,12 @@ def predict(self, batch: Batch) -> Dict[str, Tensor]:
 
         """
         user_interests = self.user_tower(batch)
+
         item_emb = self.item_tower(batch)
 
         user_emb = self.label_aware_attention(user_interests, item_emb)
-        ui_sim = self.sim(user_emb, item_emb)
+        if self._model_config.similarity == simi_pb2.Similarity.COSINE:
+            user_emb = F.normalize(user_emb, p=2.0, dim=1)
```
tiankongdeguiji: normalize we should do in tower

tiankongdeguiji (changes requested):

tzrec/models/mind.py
```diff
@@ -104,23 +154,38 @@ def forward(self, batch: Batch) -> torch.Tensor:
...
-            hist_seq_feas = self._hist_seq_mlp(grp_hist_seq)
+            if self._hist_seq_mlp_out:
+                hist_seq_feas = self._hist_seq_mlp_out(self._hist_seq_mlp(grp_hist_seq))
+            else:
+                hist_seq_feas = self._hist_seq_mlp(grp_hist_seq)
         else:
             hist_seq_feas = grp_hist_seq
 
         high_capsules = self._capsule_layer(hist_seq_feas, grp_hist_len)
+        high_capsules_mask = norm(high_capsules, dim=-1) != 0.0
```
tiankongdeguiji: we use sequence length to generate high_capsules_mask, which may be better.

tzrec/models/mind.py
```diff
@@ -242,12 +312,17 @@ def label_aware_attention(
         batch_size = user_interests.size(0)
         pos_item_emb = item_emb[:batch_size]
 
+        interest_mask = norm(user_interests, dim=-1) != 0.0
```
tiankongdeguiji: is there a better way to generate interest_mask than equal to zero?

tzrec/models/mind.py
eric-gecheng: because we need the last layer of each mlp module to have no activation

tzrec/models/mind.py
eric-gecheng: now high_capsules_mask is returned by capsule_layer module, it's caculated by sequence masking

tzrec/models/mind.py
eric-gecheng: is_inference flag is used now, so interest_mask is returned by user tower during training.

tiankongdeguiji (approved):
LGTM

PR #165, opened 2025-04-15
fix loop logic in hitrate.py

eric-gecheng:


tzrec/tools/hitrate.py
```diff
@@ -64,9 +64,8 @@ def batch_hitrate(
             )
             continue
         for interest_id in range(max_num_interests):
-            if num_interests:
-                if interest_id > num_interests[idx]:
-                    continue
+            if num_interests and interest_id >= num_interests[idx]:
+                break
```
tiankongdeguiji, line 68: why break?

tiankongdeguiji (changes requested):

tzrec/tools/hitrate.py
eric-gecheng: when interest_id is larger than the current num_interests, there is no need to check hit-rate any more. So break out the loop.

tiankongdeguiji (approved):

PR #172, opened 2025-05-01
Feature/hitrate for mind

eric-gecheng:
Notes:
1. Pick top-k items for all user interests, add duplicate item removal and sorting logics among user interests.
2. Duplicate item removal and sorting logics support both inner product and L2 distance metric.

### DIFF
--- a/examples/mind_taobao.config
+++ b/examples/mind_taobao.config
@@ -178,7 +178,7 @@ feature_configs {
         sequence_delim: "|"
         expression: "user:click_50_seq__adgroup_id"
         embedding_dim: 16
-        hash_bucket_size: 846812
+        num_buckets: 846812
     }
 }
 feature_configs {
@@ -188,7 +188,7 @@ feature_configs {
         sequence_delim: "|"
         expression: "user:click_50_seq__cate_id"
         embedding_dim: 8
-        hash_bucket_size: 12961
+        num_buckets: 12961
     }
 }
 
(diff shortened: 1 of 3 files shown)


examples/mind_taobao.config
```diff
@@ -178,7 +178,7 @@ feature_configs {
         sequence_delim: "|"
         expression: "user:click_50_seq__adgroup_id"
         embedding_dim: 16
-        hash_bucket_size: 846812
+        num_buckets: 846812
```
tiankongdeguiji, line 181: Also should fix hash_bucket_size in  docs/source/models/mind.md

tzrec/tools/hitrate.py
```diff
@@ -261,12 +271,107 @@ def batch_hitrate(
...
-        recall_distances = recall_distances.reshape(
-            [-1, args.num_interests, recall_distances.shape[-1]]
-        )
-        recall_ids = recall_ids.reshape(
-            [-1, args.num_interests, recall_distances.shape[-1]]
-        )
+
+        # In case of all-zero query vector, the corresponding knn results
+        # should be removed.
+        if args.index_type.endswith("IP"):
```
tiankongdeguiji: modify recall_distances for calculate topk candidates across all interests?
may be move the following calculation for topk candidates across all interests to a function is better.

examples/mind_taobao.config
eric-gecheng: done

tzrec/tools/hitrate.py
eric-gecheng: logics for mind model are put into a separate function now.

tiankongdeguiji (approved):

PR #194, opened 2025-06-06
[do not merge] prune mem of one shard > mem of one device for DynamicProgrammingProposer

tiankongdeguiji:
This is a plan change, merge the pr after the tzrec version bump up to 0.8.0

### DIFF
--- a/tzrec/utils/plan_util.py
+++ b/tzrec/utils/plan_util.py
@@ -212,9 +212,9 @@ class DynamicProgrammingProposer(Proposer):
         # indices of sharding_options
         self._proposal_list: List[List[int]] = []
         self._current_proposal: int = -1
-        self._plan_by_hbm = True
+        self._storage_type = "hbm"
         if not torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


tzrec/utils/plan_util.py
```diff
@@ -304,10 +301,19 @@ def feedback(
                 self._sharding_options_by_fqn.values()
             ):
                 for opt_id, sharding_option in enumerate(sharding_options):
+                    # prune mem of one shard > mem of one device
+                    if (
+                        max(
+                            [
+                                getattr(shard.storage, self._storage_type)
```
eric-gecheng, line 308: do we need to convert 
getattr(shard.storage, self._storage_type)
into numeric type such as float?

eric-gecheng (approved):

tzrec/utils/plan_util.py
tiankongdeguiji: It's a numeric type

PR #211, opened 2025-07-18
create fg.json if exist will error

chengaofei:
create fg.json if exist will error

### DIFF
--- a/tzrec/tools/create_fg_json.py
+++ b/tzrec/tools/create_fg_json.py
@@ -137,6 +137,11 @@ if __name__ == "__main__":
                     resource = o.create_resource(
                         fname, "file", file_obj=open(fpath, "rb")
                     )
+                else:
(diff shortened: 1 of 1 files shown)


tzrec/tools/create_fg_json.py
```diff
@@ -137,6 +137,11 @@
                     resource = o.create_resource(
                         fname, "file", file_obj=open(fpath, "rb")
                     )
+                else:
+                    raise ValueError(
+                        f"{fname} has already existed in project:{project}, "
+                        f"you can add '--force_update_resource true' in command"
```
tiankongdeguiji: {fname} already existed in the {project}. You can add '--force_update_resource' in the command to force update.

tiankongdeguiji (changes requested):

tiankongdeguiji (approved):

PR #213, opened 2025-07-22
Feature/fix dense embedding export in dssmv2

eric-gecheng:
Dense embedding(autodis and mlp) export fails for dssm_v2. The embedding should be splited for each tower, so the state dict should be adapted in this scenario.


tzrec/modules/dense_embedding_collection.py
```diff
@@ -190,6 +225,39 @@ def forward(self, input: Tensor) -> Tensor:
...
+    ) -> Dict[str, Any]:
+        """Override.
+
+        Split the parameter proj_w so that they can be exported when using shared
+        embedding models like dssm_v2.
+        """
+        if destination is None:
+            destination = OrderedDict()
+            destination._metadata = OrderedDict()
+        for i in range(self.proj_w.shape[0]):
```
tiankongdeguiji, line 253: We should use feature_name here rather than i.

tiankongdeguiji (changes requested):
bump up a version

tzrec/modules/dense_embedding_collection.py
eric-gecheng: resolved

tzrec/tests/configs/dssm_v2_mlpemb_fg_mock.config
```diff
@@ -0,0 +1,272 @@
...
+        embedding_name: "user_id_3_emb"
+    }
+}
+feature_configs {
+    raw_feature {
+        feature_name: "user_raw_1"
+        expression: "user:user_raw_1"
+        boundaries: [0.1, 0.2, 0.3, 0.4]
+        embedding_dim: 16
+        mlp {}
```
tiankongdeguiji, line 91: boundaries and mlp could not be used at the same time

eric-gecheng: resolved.

PR #241, opened 2025-08-12
[feat] upgrade torchrec to 1.3.0

tiankongdeguiji:


.github/workflows/benchmark.yml
```diff
@@ -8,8 +8,9 @@ on:
 jobs:
   ci-test:
     runs-on: tzrec-bench-runner
+    timeout-minutes: 960
     container:
-      image: mybigpai-public-registry.cn-beijing.cr.aliyuncs.com/easyrec/tzrec-devel:0.8
+      image: mybigpai-public-registry.cn-beijing.cr.aliyuncs.com/easyrec/tzrec-test:0.9
```
eric-gecheng: why rename the image to tzrec-test?

docker/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN sed -i "s@http://archive.ubuntu.com@http://mirrors.aliyun.com@g" /etc/apt/so
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
         build-essential ca-certificates \
-        ccache cmake gcc git vim watchman wget curl && \
+        ccache cmake gcc git vim watchman wget curl sudo iprouter2 && \
```
eric-gecheng: do we need iprouter2?

.github/workflows/benchmark.yml
tiankongdeguiji: When TorchRec releases version 1.3.0, we will proceed with building tzrec-devel version 0.9.

docker/Dockerfile
tiankongdeguiji: We need sudo and iprouter2 when training on the internal Nebula platform

tzrec/datasets/utils.py
```diff
@@ -183,12 +183,16 @@ def to(self, device: torch.device, non_blocking: bool = False) -> "Batch":
     def record_stream(self, stream: torch.Stream) -> None:
         """Record which streams have used the tensor."""
         for v in self.dense_features.values():
+            # pyre-ignore [6]
```
chengaofei, line 186: 感觉代码没问题呀，为啥要加pyfe-ignore

tiankongdeguiji: torchrec有变更，KeyedTensor入参变成了torch.cuda.streams.Stream，但Pipelinable的还是torch.Stream，先ignore

PR #262, opened 2025-09-02
[bugfix]nccl timeout

eric-gecheng:
add nccl timeout when init process group.

### DIFF
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -15,6 +15,7 @@ import json
 import os
 import shutil
 from collections import OrderedDict
+from datetime import timedelta
 from queue import Queue
 from threading import Thread
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -101,7 +102,7 @@ def init_process_group() -> Tuple[torch.device, str]:
     else:
         device: torch.device = torch.device("cpu")
         backend = "gloo"
-    dist.init_process_group(backend=backend)
+    dist.init_process_group(backend=backend, timeout=timedelta(seconds=3600))
     return device, backend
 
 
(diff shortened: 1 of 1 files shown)


tzrec/main.py
```diff
@@ -101,7 +102,7 @@ def init_process_group() -> Tuple[torch.device, str]:
     else:
         device: torch.device = torch.device("cpu")
         backend = "gloo"
-    dist.init_process_group(backend=backend)
+    dist.init_process_group(backend=backend, timeout=timedelta(seconds=3600))
```
tiankongdeguiji: use env PROCESS_GROUP_TIMEOUT_SECONDS to set timeout

eric-gecheng: done

tzrec/main.py
```diff
@@ -101,7 +102,11 @@ def init_process_group() -> Tuple[torch.device, str]:
     else:
         device: torch.device = torch.device("cpu")
         backend = "gloo"
-    dist.init_process_group(backend=backend)
+
+    nccl_timeout_sec = int(os.getenv("PROCESS_GROUP_TIMEOUT_SECONDS", "3600"))
+    dist.init_process_group(
+        backend=backend, timeout=timedelta(seconds=nccl_timeout_sec)
```
tiankongdeguiji: I think we should not modify the default timeout value
```python
pg_timeout = None
if "PROCESS_GROUP_TIMEOUT_SECONDS" in os.environ: 
    pg_timeout = timedelta(seconds=(int(os.environ("PROCESS_GROUP_TIMEOUT_SECONDS"))))
dist.init_process_group(backend=backend, timeout=pg_timeout)
```

tiankongdeguiji (changes requested):

tzrec/main.py
eric-gecheng: done

tiankongdeguiji (approved):

PR #319, opened 2025-11-14
[feat] Adapt integration test config to local cuda device count

eric-gecheng:


tzrec/tests/utils.py
```diff
@@ -59,6 +59,12 @@ def _create_random_id_data(
     return data
 
 
+def _get_nproc_per_node() -> int:
+    """Set nproc_per_node to gpu count, if cpu mode, set to 2."""
+    n_cuda = torch.cuda.device_count()
+    return n_cuda if n_cuda > 0 else 2
```
tiankongdeguiji: The test will fail when nproc_per_node > 2. https://github.com/alibaba/TorchEasyRec/blob/master/tzrec/tests/rank_integration_test.py#L114

eric-gecheng: Add env variable TEST_NPROC_PER_NODE to control.

PR #328, opened 2025-11-20
[bugfix] avoid jit convert error when using large number

eric-gecheng:
threshold = (interest_mask.float() * 2 - 1) * 1e32

however, after converting the code to TorchScript format, runtime exception occurred in jit-generated cuda kernel:
RuntimeError: default_program(22): error: extra text after expected end of number
aten_mul_1[(long long)(threadIdx.x) + 512ll * (long long)(blockIdx.x)] = (v * 2.f - 1.f) * 1.000000033181354e+32.f;

### DIFF
--- a/tzrec/models/match_model.py
+++ b/tzrec/models/match_model.py
@@ -103,7 +103,7 @@ def _sim_with_sampler(
             device=hard_neg_indices.device,
         ).to_dense()
 
-        hard_neg_ui_sim = hard_neg_ui_sim - (1 - hard_neg_mask) * float("inf")
+        hard_neg_ui_sim = hard_neg_ui_sim - (1 - hard_neg_mask) * 1e32
         return torch.cat([pos_ui_sim, neg_ui_sim, hard_neg_ui_sim], dim=-1)
 
 


tzrec/models/match_model.py
```diff
@@ -103,7 +103,7 @@ def _sim_with_sampler(
             device=hard_neg_indices.device,
         ).to_dense()
 
-        hard_neg_ui_sim = hard_neg_ui_sim - (1 - hard_neg_mask) * 1e32
+        hard_neg_ui_sim = hard_neg_ui_sim - (1 - hard_neg_mask) * float("inf")
```
tiankongdeguiji, line 106: Use 2**31 is better like https://github.com/alibaba/TorchEasyRec/commit/79f9eec649aa0efaa30a87a6ee1f97d45c5e7943#diff-7d11485e2f2caa22304da371501fa75d76a4ab58494b1e52c9da364d23b36914, will 'inf' raise training error?

eric-gecheng: no, here float('inf') is in an expression, it is taken by pytorch as a tensor, the softmax will give 0. But if directly assign tensor values to float('inf') will cause nan after softmax. let me check 10**32, it's another way.

PR #333, opened 2025-11-26
[feat]add assert to avoid using ckpt predict for two tower models

eric-gecheng:
Two-tower match models need export to split the graph. Avoid model predict by checkpoint directly.

### DIFF
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -1214,12 +1214,8 @@ def predict_checkpoint(
 
     pipeline_config = config_util.load_pipeline_config(pipeline_config_path)
 
-    assert (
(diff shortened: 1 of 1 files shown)


tzrec/main.py
```diff
@@ -1213,6 +1213,18 @@ def predict_checkpoint(
         output_cols = [x.strip() for x in output_columns.split(",")]
 
     pipeline_config = config_util.load_pipeline_config(pipeline_config_path)
+
+    assert (
```
tiankongdeguiji: Why does predict_checkpoint not support dssm?

eric-gecheng: because two tower models need to split the graph into user/item parts before predict.

PR #525, opened 2026-05-24
[feat] distributed embedding export 

eric-gecheng:


tzrec/utils/checkpoint_util.py
```diff
@@ -124,6 +142,26 @@ def create_local_plan(self) -> LoadPlan:
                         "[{meta_fqn}], will be deprecated when tzrec version >= 1.0.0"
                     )
 
+            # INPUT_TILE=3 export adds user-side twin modules (`ebc_user`,
```
tiankongdeguiji: simplify all comments in the pr

tzrec/utils/checkpoint_util.py
```diff
@@ -124,6 +142,26 @@ def create_local_plan(self) -> LoadPlan:
...
+            # `mc_ebc_user`, `ec_dict_user`, `mc_ec_dict_user`) whose
+            # state_dict keys do not exist in the training checkpoint
+            # (training never sets INPUT_TILE). Fall back to the non-user
+            # counterpart that does exist. Equivalent to the explicit
+            # `emb_ckpt_mapping.txt` generated in `export_model_normal`,
+            # but works without the call site pre-computing a mapping.
+            if meta_fqn not in self.metadata.state_dict_metadata:
```
tiankongdeguiji: Is `emb_ckpt_mapping.txt` no longer needed?

tzrec/utils/checkpoint_util.py
tiankongdeguiji: Enable it only when `INPUT_TILE=3`.

tzrec/utils/checkpoint_util.py
```diff
@@ -872,16 +954,36 @@ def restore_model(
...
+            # Training never sets INPUT_TILE, but exporting with
+            # INPUT_TILE=3 adds twin user-side modules (`ebc_user`,
+            # `mc_ebc_user`, `ec_dict_user`, `mc_ec_dict_user`) that
+            # share dynamic-embedding tables with their non-user counterparts.
+            # Build a local symlink view instead of mutating the checkpoint
+            # directory, which may be read-only or remote-mounted.
+            input_tile = os.environ.get("INPUT_TILE", "")
+            dynamicemb_load_path = dynamicemb_path
+            dynamicemb_view = None
+            if input_tile.startswith("3"):
```
tiankongdeguiji: acc_utils.is_input_tile_emb

tzrec/utils/checkpoint_util.py
```diff
@@ -872,16 +954,36 @@ def restore_model(
...
+                dynamicemb_view = tempfile.TemporaryDirectory(
```
tiankongdeguiji, line 963: Why should each rank create its own separate `dynamicemb_view`?

tzrec/utils/checkpoint_util.py
```diff
@@ -794,6 +832,50 @@ def _a2a(
...
+    for entry in os.listdir(dynamicemb_path):
+        full_path = os.path.join(dynamicemb_path, entry)
+        if not os.path.isdir(full_path):
+            continue
+        entries.append((entry, full_path))
+        link_path = os.path.join(view_path, entry)
+        if not os.path.lexists(link_path):
+            os.symlink(full_path, link_path, target_is_directory=True)
+
+    for entry, full_path in entries:
```
tiankongdeguiji, line 856: Can we reuse `_INPUT_TILE_USER_REPLACEMENTS` here and replace `old_pat` with `new_pat` in the path?

tzrec/utils/export_util.py
```diff
@@ -77,6 +79,29 @@
...
+def ensure_input_tile_for_distributed_embedding() -> None:
+    """Ensure distributed embedding export uses INPUT_TILE=3."""
+    if not env_util.use_distributed_embedding():
+        return
+
+    # Distributed embedding export only supports INPUT_TILE=3.
+    # Default to 3 when unset; warn and override if set to anything else.
+    current_input_tile = os.environ.get("INPUT_TILE")
+    if current_input_tile is None:
+        os.environ["INPUT_TILE"] = "3"
```
tiankongdeguiji, line 92: Why is only `INPUT_TILE=3` supported? If it is the only supported value, should we always log `INPUT_TILE=3`?

tzrec/utils/export_util.py
```diff
@@ -92,8 +117,14 @@ def export_model(
     input path; falls back to `pipeline_config.train_input_path` when None.
     """
     use_rtp = env_util.use_rtp()
-
-    impl = export_rtp_model if use_rtp else export_model_normal
+    ensure_input_tile_for_distributed_embedding()
```
tiankongdeguiji: do it in export_distributed_embedding function

tzrec/utils/export_util.py
```diff
@@ -317,6 +348,36 @@ def export_model_normal(
                 shutil.copy(asset, save_dir)
 
 
+def _prepare_single_rank_distributed_embedding_export() -> bool:
```
tiankongdeguiji, line 371: Why do we need `_prepare_single_rank_distributed_embedding_export`?

tzrec/main.py
```diff
@@ -977,6 +980,8 @@ def export(
     """
     is_rank_zero = int(os.environ.get("RANK", 0)) == 0
 
+    ensure_input_tile_for_distributed_embedding()
```
tiankongdeguiji, line 983: Why do we need to call `ensure_input_tile_for_distributed_embedding` again here?

tzrec/utils/export_util.py
```diff
@@ -1204,3 +1284,804 @@ def _seq_feat_name(seq_name: str) -> str:
...
+    if not os.path.exists(save_dir_sparse):
+        os.makedirs(save_dir_sparse)
+    local_tensor_path = os.path.join(save_dir_sparse, f"{local_tensor_name}.npz")
+    logger.info(f"save sparse tensors to {local_tensor_path}")
+    # np.savez(local_tensor_path, **local_tensor)
+    import tempfile
+
+    # OSS mounted file system may have problem in file seek, so first
+    # save to a temp file then move to target path
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".npz") as f:
```
tiankongdeguiji, line 1469: For very large embeddings, the temporary directory may run out of storage space. Also, this operation is not needed for NAS or CPFS.

tzrec/utils/export_util.py
```diff
@@ -1204,3 +1284,804 @@ def _seq_feat_name(seq_name: str) -> str:
...
+        return candidates[0]
+    if not candidates:
+        raise KeyError(f"sparse embedding {emb_name} is not in export metadata")
+    raise ValueError(
+        f"sparse embedding {emb_name} appears in multiple sparse collection "
+        f"kinds; cannot resolve export name without role, got role={role}"
+    )
+
+
+def _get_sparse_embedding_tensor(
```
tiankongdeguiji, line 1785: Can we reuse `_get_rtp_embedding_tensor` here? What is the difference between the two implementations?

Also, I’m concerned that the following code may cause OOM for very large embeddings:

```python
torch.tensor(np.fromfile(f, dtype=np.int64), dtype=torch.int64)
```

The dynamic embedding table is loaded and stored by chunk, so we may need to avoid loading the whole file at once.

tzrec/utils/export_util.py
```diff
@@ -1204,3 +1284,804 @@ def _seq_feat_name(seq_name: str) -> str:
...
+        temp_path = f.name
+    shutil.move(temp_path, local_tensor_path)
+
+    if dynamic_local_tensor:
+        dynamic_tensor_name = f"sparse_dynamic_embedding-{rank:02d}-of-{world_size:02d}"
+        dynamic_tensor_path = os.path.join(
+            save_dir_sparse, f"{dynamic_tensor_name}.npz"
+        )
+        logger.info(f"save dynamic sparse tensors to {dynamic_tensor_path}")
+        with tempfile.NamedTemporaryFile(delete=False, suffix=".npz") as f:
```
tiankongdeguiji, line 1480: I think using a zip file and streaming the embeddings into it as NumPy arrays chunk by chunk would be better. This avoids materializing the full embedding tensor in memory.

For example:

```python
import zipfile

import numpy as np
from numpy.lib import format as npformat


def write_npy_in_chunks(zf, name, chunks, shape, dtype):
    dtype = np.dtype(dtype)

    with zf.open(name, mode="w") as f:
        # Write the .npy header.
        npformat.write_array_header_2_0(
            f,
            {
                "descr": npformat.dtype_to_descr(dtype),
                "fortran_order": False,
                "shape": tuple(shape),
            },
        )

        written = 0
        for chunk in chunks:
            chunk = np.asarray(chunk, dtype=dtype)

            # Assume chunks are split along the first dimension.
            if chunk.shape[1:] != tuple(shape[1:]):
                raise ValueError(f"Bad chunk shape: {chunk.shape}")

            written += chunk.shape[0]
            if written > shape[0]:
                raise ValueError("Too many rows written")

            f.write(np.ascontiguousarray(chunk).tobytes(order="C"))

        if written != shape[0]:
            raise ValueError(f"Expected {shape[0]} rows, got {written}")


def chunks_a():
    for _ in range(10):
        yield np.random.rand(100, 32).astype(np.float32)


def chunks_b():
    for _ in range(5):
        yield np.random.randint(0, 100, size=(200,), dtype=np.int64)


with zipfile.ZipFile(
    "arrays.zip",
    "w",
    compression=zipfile.ZIP_DEFLATED,
    compresslevel=6,
) as zf:
    write_npy_in_chunks(
        zf,
        "a.npy",
        chunks_a(),
        shape=(1000, 32),
        dtype=np.float32,
    )

    write_npy_in_chunks(
        zf,
        "b.npy",
        chunks_b(),
        shape=(1000,),
        dtype=np.int64,
    )
```

tzrec/utils/export_util.py
```diff
@@ -1204,3 +1284,804 @@ def _seq_feat_name(seq_name: str) -> str:
...
+
+        # Save FG
+        if not has_fg_asset:
+            logger.info("saving fg json...")
+            fg_json = create_fg_json(features, asset_dir=save_dir)
+            fg_json["features"].extend(additional_fg)
+            with open(os.path.join(save_dir, "fg.json"), "w") as f:
+                json.dump(fg_json, f, indent=4)
+
+    dist.barrier()
```
tiankongdeguiji: If only single-rank export is supported, do we still need these?

If so, we should avoid the file-based approach and use `dist.gather_object` instead, because the distributed file system may not synchronize in time.

tiankongdeguiji (changes requested):

tiankongdeguiji:
bump up version to 1.3.1

tzrec/utils/checkpoint_util.py
eric-gecheng: when exporting a model with USE_DISTRIBUTED_EMBEDDING=1, INPUT_TILE=3 is enabled automatically.

tzrec/utils/checkpoint_util.py
eric-gecheng: PartialLoadPlanner is doing remap

tzrec/utils/export_util.py
eric-gecheng: noted, will be implemented in next pr.

tzrec/utils/export_util.py
eric-gecheng: will be implemented in next pr

tzrec/utils/checkpoint_util.py
eric-gecheng: will migrate to PartialLoadPlanner in next PR

tzrec/utils/checkpoint_util.py
eric-gecheng: solved.

tzrec/main.py
eric-gecheng: Solved. Removed ensure_input_tile_for_distributed_embedding in export_model function.

tzrec/utils/export_util.py
eric-gecheng: INPUT_TILE=3 is the most efficient way for processor inference in distributed embedding setup, so that nvembedding will only do lookup once.

tzrec/utils/export_util.py
eric-gecheng: removed the function here, it will be called in export function in main.py.

tzrec/utils/export_util.py
eric-gecheng: force single rank export

tzrec/utils/export_util.py
eric-gecheng: optimize in next PR

tzrec/utils/export_util.py
eric-gecheng: dist.barrier() removed in current single-rank export scenario.

tzrec/utils/checkpoint_util.py
eric-gecheng: separate dynamicemb_view is safer when doing cleanup.

tiankongdeguiji (approved):

PR #539, opened 2026-06-05
[feat] SID: add SidRqkmeans model (FAISS-trained residual K-Means)

WhiteSwan1:
Second of three PRs splitting the Semantic-ID models onto the shared base from #538. Adds the concrete RQ-KMeans backend on top of ResidualQuantizer / BaseSidModel; RQ-VAE follows in PR3.

- tzrec/modules/sid/kmeans.py: KMeansLayer centroid container + recon_diagnostics.
- tzrec/modules/sid/residual_kmeans_quantizer.py: ResidualKMeansQuantizer (FAISS-trained, FX-traceable forward, non-uniform per-layer codebooks).
- tzrec/models/sid_rqkmeans.py: SidRqkmeans(BaseSidModel) - gradient -free; reservoir-samples embeddings during the train loop and fits FAISS once in on_train_end.
- tzrec/models/model.py: BaseModel.on_train_end() no-op lifecycle hook.
- tzrec/main.py: invoke on_train_end after the train loop and force the tail checkpoint so post-hook state is persisted.
- protos: SidRqkmeans message + ModelConfig registration (601; 600 is reserved for SidRqvae in PR3).
- tests: kmeans_test, ResidualKMeansQuantizerTest, sid_rqkmeans_test.

### DIFF
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -502,12 +502,8 @@ def _train_and_evaluate(
 
     # One-shot end-of-loop hook (default no-op). Some models do real work
     # here — e.g. SidRqkmeans fits its FAISS codebook from the embeddings
-    # collected during training. Since that mutates model state, force the
-    # tail-save below to fire so the post-hook state is persisted even when
-    # the last in-loop checkpoint coincided with the final step.
+    # collected during training.
     _model.on_train_end()
-    if last_ckpt_step == i_step:
-        last_ckpt_step = -1
 
     _log_train(
         i_step,
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -329,7 +329,7 @@ def _train_and_evaluate(
     eval_result_filename: str = TRAIN_EVAL_RESULT_FILENAME,
     check_all_workers_data_status: bool = False,
     ignore_restore_optimizer: bool = False,
-    dataloader_state: Optional[Dict[str, int]] = None,
+    dataloader_state: Optional[Dict[str, Any]] = None,
 ) -> None:
     """Train and evaluate the model."""
     is_rank_zero = int(os.environ.get("RANK", 0)) == 0
@@ -355,6 +355,16 @@ def _train_and_evaluate(
         save_checkpoints_epochs = train_config.save_checkpoints_epochs
     else:
         save_checkpoints_steps = train_config.save_checkpoints_steps
+
+    # The CheckpointManager owns the save cadence + dedupe + watermark; the loop
+    # only gathers per-worker event-times and runs eval after a save.
+    ckpt_manager.set_save_policy(
+        save_checkpoints_steps,
+        save_checkpoints_epochs,
+        train_config.save_checkpoints_timestamp_interval,
+        list(train_config.save_checkpoints_timestamps),
+        train_config.save_checkpoints_timestamp_quorum,
+    )
 
     plogger = None
     summary_writer = None
@@ -395,10 +405,27 @@ def _train_and_evaluate(
         )
         prof.start()
 
-    last_ckpt_step = -1
     i_step = 0
     i_epoch = 0
     losses = {}
+
+    def run_eval(step: int, epoch: int) -> None:
+        """Run eval after a checkpoint save, if an eval dataloader is configured."""
+        if eval_dataloader is not None:
+            _evaluate(
+                model,
+                eval_dataloader,
+                eval_config,
+                eval_result_filename=eval_result_filename,
+                global_step=step,
+                eval_summary_writer=eval_summary_writer,
+                global_epoch=epoch,
+                check_all_workers_data_status=check_all_workers_data_status,
+            )
+            model.train()
+
+    # this rank's last consumed event-time, reused by the epoch / final saves
+    data_timestamp = -1.0
     for i_epoch in epoch_iter:
         pipeline = create_train_pipeline(
             model,
@@ -457,41 +484,29 @@ def _train_and_evaluate(
                 i_step -= 1
                 break
 
-            if save_checkpoints_steps > 0 and i_step > 0:
-                if i_step % save_checkpoints_steps == 0:
-                    last_ckpt_step = i_step
-                    ckpt_manager.save(i_step, model, optimizer, dataloader_state)
-                    if eval_dataloader is not None:
-                        _evaluate(
-                            model,
-                            eval_dataloader,
-                            eval_config,
-                            eval_result_filename=eval_result_filename,
-                            global_step=i_step,
-                            eval_summary_writer=eval_summary_writer,
-                            global_epoch=i_epoch,
-                            check_all_workers_data_status=check_all_workers_data_status,
-                        )
-                        model.train()
+            # Single entry point for step / epoch / event-time saves + dedupe;
+            # maybe_save reconciles this rank's event-time across ranks in lockstep.
+            data_timestamp = batch.data_timestamp
+            if ckpt_manager.maybe_save(
+                i_step,
+                model,
+                optimizer,
+                dataloader_state,
+                data_timestamp=data_timestamp,
+            ):
+                run_eval(i_step, i_epoch)
             if train_config.is_profiling:
                 prof.step()
 
-        if save_checkpoints_epochs > 0 and i_step > 0:
-            if (i_epoch + 1) % save_checkpoints_epochs == 0:
-                last_ckpt_step = i_step
-                ckpt_manager.save(i_step, model, optimizer, dataloader_state)
-                if eval_dataloader is not None:
-                    _evaluate(
-                        model,
-                        eval_dataloader,
-                        eval_config,
-                        eval_result_filename=eval_result_filename,
-                        global_step=i_step,
-                        eval_summary_writer=eval_summary_writer,
-                        global_epoch=i_epoch,
-                        check_all_workers_data_status=check_all_workers_data_status,
-                    )
-                    model.train()
+        if ckpt_manager.maybe_save(
+            i_step,
+            model,
+            optimizer,
+            dataloader_state,
+            epoch=i_epoch,
+            data_timestamp=data_timestamp,
+        ):
+            run_eval(i_step, i_epoch)
 
         if use_step and i_step >= train_config.num_steps - 1:
             break
@@ -518,20 +533,19 @@ def _train_and_evaluate(
         summary_writer.close()
     if train_config.is_profiling:
         prof.stop()
-    if last_ckpt_step != i_step or is_ckpt_after_train:
-        ckpt_manager.save(i_step, model, optimizer, dataloader_state)
-        if eval_dataloader is not None:
-            _evaluate(
-                model,
-                eval_dataloader,
-                eval_config,
-                eval_result_filename=eval_result_filename,
-                global_step=i_step,
-                eval_summary_writer=eval_summary_writer,
-                global_epoch=i_epoch,
-                check_all_workers_data_status=check_all_workers_data_status,
-            )
-            model.train()
+    # ``force`` re-fires the save past maybe_save's per-step dedupe when
+    # on_train_end mutated persistable state (e.g. SidRqkmeans fit its codebook)
+    # after the last in-loop save landed on the final step.
+    if ckpt_manager.maybe_save(
+        i_step,
+        model,
+        optimizer,
+        dataloader_state,
+        data_timestamp=data_timestamp,
+        final=True,
+        force=is_ckpt_after_train,
+    ):
+        run_eval(i_step, i_epoch)
     ckpt_manager.close()
 
 
@@ -633,7 +647,7 @@ def train_and_evaluate(
                 )
 
     # Restore dataloader checkpoint state
-    dataloader_state: Optional[Dict[str, int]] = None
+    dataloader_state: Optional[Dict[str, Any]] = None
     if ckpt_path and continue_train:
         dataloader_state = ckpt_manager.restore_dataloader_state(ckpt_path)
         if dataloader_state:
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -500,11 +500,9 @@ def _train_and_evaluate(
             if lr.by_epoch:
                 lr.step()
 
-    # One-shot end-of-loop hook (default no-op). Some models do real work
-    # here — e.g. SidRqkmeans fits its FAISS codebook from the embeddings
-    # collected during training. When the hook mutated state that must be
-    # persisted, it returns True so the tail save below fires even if the
-    # last in-loop checkpoint already landed on the final step.
+    # One-shot end-of-loop hook (default no-op; e.g. SidRqkmeans fits its FAISS
+    # codebook here). Returns True if it mutated persistable state, forcing the
+    # tail save below even when the last in-loop checkpoint hit the final step.
     is_ckpt_after_train = _model.on_train_end()
 
     _log_train(
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -500,6 +500,15 @@ def _train_and_evaluate(
             if lr.by_epoch:
                 lr.step()
 
+    # One-shot end-of-loop hook (default no-op). Some models do real work
+    # here — e.g. SidRqkmeans fits its FAISS codebook from the embeddings
+    # collected during training. Since that mutates model state, force the
+    # tail-save below to fire so the post-hook state is persisted even when
+    # the last in-loop checkpoint coincided with the final step.
+    _model.on_train_end()
+    if last_ckpt_step == i_step:
+        last_ckpt_step = -1
+
     _log_train(
         i_step,
         losses,
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -516,9 +516,10 @@ def _train_and_evaluate(
                 lr.step()
 
     # One-shot end-of-loop hook (default no-op; e.g. SidRqkmeans fits its FAISS
-    # codebook here). Returns True if it mutated persistable state, forcing the
-    # tail save below even when the last in-loop checkpoint hit the final step.
-    is_ckpt_after_train = _model.on_train_end()
+    # codebook here). SID models run with periodic checkpointing disabled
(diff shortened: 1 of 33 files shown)


tzrec/models/sid_rqkmeans.py
```diff
@@ -0,0 +1,349 @@
...
+        local = self._reservoir_sample()
+        self._reset_reservoir()
+
+        if is_ddp:
+            # Each rank ships its (capped) reservoir to rank0, which fits.
+            rank = dist.get_rank()
+            gathered: Optional[List[Optional[torch.Tensor]]] = (
+                [None] * dist.get_world_size() if rank == 0 else None
+            )
+            dist.gather_object(local, gathered, dst=0)
```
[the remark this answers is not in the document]
WhiteSwan1: thanks for your comment. however, `get_dist_object_pg()` returns `dist.GroupMember.WORLD` (the NCCL group) whenever `WORLD.size() == world_size`, i.e. every normal launch — not a gloo group. So `gather_object(..., group=get_dist_object_pg())` is the same NCCL group as the default and changes nothing; the cited `odps_dataset.py:791` likewise gathers on WORLD/NCCL in the common case (its `self._pg = get_dist_object_pg(world_size)` defaults to WORLD). `gather_object` works here because `init_process_group` sets the per-rank device, and the reservoir is bounded by `train_sample_size`. i think no change needed.

tzrec/modules/sid/residual_kmeans_quantizer.py
```diff
@@ -0,0 +1,241 @@
...
+        x0 = x.clone() if verbose else None
+
+        # Use FAISS GPU compute when a faiss-gpu build is present; an explicit
+        # ``gpu`` in faiss_kmeans_kwargs always wins. NB faiss reads ``gpu`` as a
+        # GPU *count* (1 = one GPU = the current/rank0 device), not a device
+        # index — passing an index of 0 is falsy and silently falls back to CPU.
+        kwargs = dict(self.faiss_kmeans_kwargs)
+        if "gpu" not in kwargs:
+            kwargs["gpu"] = (
+                1 if (faiss.get_num_gpus() > 0 and torch.cuda.is_available()) else False
```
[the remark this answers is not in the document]
WhiteSwan1: thanks, this is exactly right. `gpu` is a count, `1 == True` collapses to `get_num_gpus()` (all GPUs), and a device *index* of `0` is falsy → silent CPU. Since faiss's count kwarg genuinely can't pin a rank0-only fit to a single device (1 → all; N≥2 → N devices from 0), I dropped the index/count approach entirely.

Fixed in `2bb5abc`: `train_offline` now defaults to a CPU fit (`kwargs.setdefault("gpu", False)`). The fit is a bounded one-shot over the reservoir subsample (`max(K) * max_points_per_centroid`), so CPU is cheap, and it avoids sharding faiss memory onto the other ranks' GPUs. An explicit `gpu` in `faiss_kmeans_kwargs` still opts into GPU. Also added a log line in `33acbe6` announcing the fit device so the CPU default isn't silent.

(Side note this surfaced: the original `gpu=torch.cuda.current_device()` meant the GPU path was effectively dead — `current_device()==0`→CPU — so this also fixes the silent-CPU-fallback flagged earlier.)

tzrec/main.py
```diff
@@ -513,7 +518,7 @@ def _train_and_evaluate(
         summary_writer.close()
     if train_config.is_profiling:
         prof.stop()
-    if last_ckpt_step != i_step:
+    if last_ckpt_step != i_step or is_ckpt_after_train:
```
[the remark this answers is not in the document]
WhiteSwan1: Agreed it's the load-bearing line for a FAISS-only model. It's currently covered indirectly: `test_post_fit_checkpoint_round_trips` verifies the fitted centroids round-trip through `state_dict`, and the `on_train_end() -> is_ckpt_after_train` contract is unit-tested (`True` on a real fit, `False` on empty). A full `train_and_evaluate` integration test for the save-ordering is heavier; can add it as a follow-up if you'd prefer it in this PR.

tzrec/models/sid_rqkmeans.py
```diff
@@ -0,0 +1,360 @@
...
+
+        codes, quantized = self._quantizer(embedding)
+
+        predictions: Dict[str, torch.Tensor] = {
+            "codes": codes,
+        }
+
+        if self.is_eval:
+            predictions["quantized"] = quantized
+            predictions["input_embedding"] = embedding
```
tiankongdeguiji: why input_embedding need to be put into predictions?

tzrec/models/sid_rqkmeans_test.py
```diff
@@ -0,0 +1,535 @@
...
+    for i, p in enumerate(procs):
+        p.join(timeout=timeout)
+        if p.is_alive():
+            p.terminate()
+            raise RuntimeError(f"worker-{i} deadlocked (timed out after {timeout}s).")
+        if p.exitcode != 0:
+            raise RuntimeError(f"worker-{i} failed (exitcode={p.exitcode}).")
+
+
+class SidRqkmeansDistTest(unittest.TestCase):
```
tiankongdeguiji: Put integration test into test/sid_integration_test.py like other integration_test

tzrec/models/sid_rqkmeans.py
```diff
@@ -0,0 +1,360 @@
...
+        Args:
+            predictions (dict): a dict of predicted result.
+            batch (Batch): input batch data.
+
+        Return:
+            losses (dict): a dict of loss tensor.
+        """
+        return {"dummy_loss": self._dummy_param.sum() * 0.0}
+
+    def init_metric(self) -> None:
```
tiankongdeguiji: The init_metric and update_metric logic should be moved to BaseSidModel. Add a custom rel_loss metric, such as MeanAbsolutePercentageError, in torchmetrics.

tzrec/models/sid_rqkmeans.py
```diff
@@ -0,0 +1,360 @@
...
+            into a coordinated ``RuntimeError``).
+        """
+        is_ddp = (
+            dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1
+        )
+
+        local = self._reservoir_sample()
+        self._reset_reservoir()
+
+        if is_ddp:
```
tiankongdeguiji: do not need to support ddp.

tzrec/modules/sid/kmeans.py
```diff
@@ -0,0 +1,175 @@
...
+        """Assign points to nearest centroid.
+
+        Args:
+            batch (Tensor): data points, shape (B, D).
+
+        Returns:
+            Tensor: cluster indices, shape (B,).
+        """
+        dists = _squared_euclidean_distance(batch, self.centroids)
+        return torch.argmin(dists, dim=-1)
```
tiankongdeguiji: use `torch.cdist(x, centroids).argmin(-1)`, do not need _squared_euclidean_distance

tiankongdeguiji (changes requested):

tzrec/modules/sid/residual_quantizer_test.py
```diff
@@ -14,6 +14,9 @@
 import torch
```
tiankongdeguiji, line 14: Rename file to residual_kmeans_quantizer.py

tzrec/modules/sid/residual_kmeans_quantizer.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        # CPU-only: SidRqkmeans refuses to init when CUDA is visible, but this
+        # quantizer is a standalone module — assert the host-tensor contract it
+        # relies on so misuse fails here, not deep inside faiss.
+        assert not inputs.is_cuda, "train_offline is CPU-only; got a CUDA tensor"
+        assert inputs.dim() == 2 and inputs.shape[1] == self.embed_dim, (
+            f"inputs must be (N, {self.embed_dim}), got {tuple(inputs.shape)}"
+        )
+        # Own one contiguous float32 copy to update in place as the residual.
+        x = inputs.detach().to(dtype=torch.float32, copy=True).contiguous()
```
tiankongdeguiji: why need to copy=True?

tzrec/modules/sid/kmeans.py
```diff
@@ -0,0 +1,269 @@
...
+        return self._buf[: self._n_filled]
+
+    def reset(self) -> None:
+        """Drop the buffer and counters to free host memory."""
+        self._buf = None
+        self._n_filled = 0
+        self._n_seen = 0
+
+
+class KMeansLayer(nn.Module):
```
tiankongdeguiji: Add a QuantizeLayer base class, such as
```
class QuantizeLayer(nn.Module):
    def quantize(
        self,
        x: torch.Tensor,
        temperature: float = 1.0,
    ) -> QuantizeOutput:

    def lookup(self, ids: torch.Tensor) -> torch.Tensor:

    def get_codebook_embeddings(self) -> torch.Tensor:
```
so that KMeansQuantizeLayer and VectorQuantizeLayer can use it.

tzrec/protos/models/sid_model.proto
```diff
@@ -0,0 +1,32 @@
...
+    // may differ per layer (non-uniform codebooks such as [256, 512, 1024]
+    // are supported — the FAISS backend fits a separate ``faiss.Kmeans``
+    // per layer).
+    repeated uint32 codebook = 3;
+    // L2-normalize residuals before each layer.
+    optional bool normalize_residuals = 4 [default = false];
+    // Extra kwargs forwarded to faiss.Kmeans(D, K, **kwargs) as a
+    // loosely-typed dict, e.g. {niter: 20, gpu: true, verbose: true,
+    // spherical: false, seed: 1234}.
+    optional google.protobuf.Struct faiss_kmeans_kwargs = 5;
```
tiankongdeguiji: Add a proto message for faiss_kmeans_kwargs, list the kwargs in it. Faiss needs strictly typed kwargs, _coerce_proto_numbers is not safe

tzrec/main.py
```diff
@@ -528,13 +533,17 @@ def run_eval(step: int, epoch: int) -> None:
...
+    # on_train_end mutated persistable state (e.g. SidRqkmeans fit its codebook)
+    # after the last in-loop save landed on the final step.
     if ckpt_manager.maybe_save(
         i_step,
         model,
         optimizer,
         dataloader_state,
         data_timestamp=data_timestamp,
         final=True,
+        force=is_ckpt_after_train,
```
tiankongdeguiji: Do not need to add the`force` param, set save_checkpoints_epochs = save_checkpoints_steps = 0 in config.

tiankongdeguiji (changes requested):

tzrec/main.py
```diff
@@ -515,6 +515,12 @@ def run_eval(step: int, epoch: int) -> None:
             if lr.by_epoch:
                 lr.step()
 
+    # One-shot end-of-loop hook (default no-op; e.g. SidRqkmeans fits its FAISS
+    # codebook here). SID models run with periodic checkpointing disabled
+    # (save_checkpoints_steps/epochs = 0), so the tail final=True save below is
+    # the only checkpoint and persists whatever on_train_end produced.
+    _model.on_train_end()
```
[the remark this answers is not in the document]
WhiteSwan1: Acknowledged, and the analysis is correct — with periodic checkpointing enabled  and the last in-loop save landing on the final step, the tail `final=True` save  is deduped and the fitted codebook is lost.
  
  We're declining the enforcement change for this PR: per the maintainer's  guidance we rely on the config convention (`save_checkpoints_steps = 0`,  `save_checkpoints_epochs = 0`) rather than adding save-path machinery, and we'd  prefer not to expand `main.py`'s save logic in this PR. We've added a TODO in `SidRqkmeans.on_train_end` to harden the save logic (enforce the contract or  bypass the dedupe for this final save) in a future update.

tzrec/models/sid_rqkmeans.py
```diff
@@ -0,0 +1,198 @@
...
+        features: List[BaseFeature],
+        labels: List[str],
+        sample_weights: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(model_config, features, labels, sample_weights, **kwargs)
+
+        # CPU-only: embeddings, reservoir, and FAISS fit all stay on the host,
+        # so there are no device copies. Refuse to run when CUDA is visible.
+        if torch.cuda.is_available():
```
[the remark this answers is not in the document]
WhiteSwan1: Intentional for v1. The `SidRqkmeans` implementation convention deliberately  restricts the **entire model — both training and inference — to a single CPU  process**, not just the FAISS fit. So the `__init__`-scoped guards are by  design: every entry point (train / eval / export / predict) constructs the  model, and we want all of them to fail fast and uniformly the moment CUDA is  visible or `world_size > 1`, rather than silently running a path we haven't  validated for GPU/multi-process in this version.

  You're right that post-fit nearest-centroid assignment is read-only and  embarrassingly parallel, and that a GPU/multi-process serving path for  corpus-wide SID assignment is a real optimization. We're scoping that out of v1  on purpose — keeping one CPU-only execution model end-to-end keeps the  device/process semantics simple and the codebook bit-exact across fit and  assignment. Lifting the constraint to a fit-scoped guard (and making the  reservoir/assignment device-agnostic) is a sensible follow-up once we need the  throughput; we'd rather not widen the supported surface area before it's  exercised and tested.

tzrec/modules/sid/residual_kmeans_quantizer.py
```diff
@@ -0,0 +1,264 @@
...
+        N = x.shape[0]
+        # Fail loudly on a too-small corpus: faiss.Kmeans only warns (not
+        # errors) when N < K and returns a degenerate codebook, which the
+        # all-zero poison guard in KMeansQuantizeLayer would not catch.
+        max_k = max(self.n_embed_list)
+        if N < max_k:
+            raise RuntimeError(
+                f"need >= {max_k} points to fit the codebook (largest layer K), "
+                f"got N={N}"
+            )
```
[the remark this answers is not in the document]
WhiteSwan1: Thanks for the response! We handle NaN prevention during upstream sample generation, so this shouldn't be an issue.

tzrec/models/sid_model.py
```diff
@@ -86,7 +96,17 @@ def _extract_feature(
...
+        embedding = kt[feature_name]
+        # Guard a misconfigured feature width: a (B, 1) tensor (raw_feature
+        # missing value_dim, which defaults to 1) would otherwise broadcast
+        # silently downstream and fit a degenerate rank-1 codebook.
+        if embedding.dim() != 2 or embedding.shape[1] != self._input_dim:
+            raise ValueError(
+                f"feature '{feature_name}' has shape {tuple(embedding.shape)}, "
+                f"expected (B, {self._input_dim}); check that its value_dim "
+                "matches the SID input_dim."
+            )
```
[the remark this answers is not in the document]
WhiteSwan1: Thanks for your reply, but the embedding cannot be [B, 1]. To keep the code concise, we'll skip this change in the current PR.

WhiteSwan1:
Thanks for both github-actions passes — went through them all. Dispositions:


**Fixed:**
- proto wording + `codebook≥1`/`input_dim≥1` validation; 
- removed the duplicated relative-L1 helper (the metric is now the single source);
- float64/long `RelativeL1` accumulators;
-  `assert→raise` for the `-O`-stripped data-corruption guards; 
- mock config `save_checkpoints=0`; 
- corrected the faiss-N<K comment; 
- guard message now says `CUDA_VISIBLE_DEVICES="-1"`; 
- added negative tests for every fail-fast path and strengthened the integration test to assert finite reconstruction metrics.


**Declined / deferred (with reason):**
  - `__init__` CPU/world_size guards stay in `__init__` — v1 is intentionally CPU-only end to end (training *and* inference), so failing fast at every entry point is the intended contract. Because KMeans inference is so simple and only involves nearest-neighbor search, it's unlikely we'll need GPU inference down the road.
  - Checkpoint-dedupe enforcement — kept the `save_checkpoints=0` convention as agreed; left a TODO to harden it in a follow-up.
  - ReservoirSampler `(B,1)` width guard — the embedding width is never 1 in practice, so reverted.
  - Empty-vs-small-reservoir consistency and a NaN/Inf input check — reasonable, but behavior/logic changes better suited to a follow-up.
  
  Also fixed the CPU-only end-to-end test that was failing on the GPU CI image: it now runs on the CPU CI job and skips where CUDA is present.

WhiteSwan1:
Thanks @tiankongdeguiji — all addressed on latest (`5f5af01`):

**Round 1**
1. `input_embedding` in predictions — removed; `predict` exposes only the reconstruction `x_hat` (eval, once fitted), and `update_metric` re-extracts the target from the batch.
2. Integration test — moved to `tzrec/tests/sid_integration_test.py` (with a mock config); it also asserts the post-fit eval reports finite metrics.
3. `init_metric`/`update_metric` → `BaseSidModel`; `rel_loss` is now a torchmetric — a **symmetric** `RelativeL1` rather than `MeanAbsolutePercentageError` (MAPE's `|t-p|/|t|` is asymmetric and unbounded as target→0; the symmetric form matches OpenOneRec's `calc_loss`). Happy to switch to the builtin if you prefer.
4. DDP — removed; CPU-only, single-process, with a `world_size>1` fail-fast in `__init__`.
5. `torch.cdist(x, centroids).argmin(-1)` — adopted; `_squared_euclidean_distance` deleted.

**Round 2**
1. File renamed to `residual_kmeans_quantizer.py`.
2. `copy=True` — removed; `train_offline` now explicitly *consumes* its input (documented), and the caller decides whether to copy.
3. Added a `QuantizeLayer` base class (`tzrec/modules/sid/quantize_layer.py`); `KMeansQuantizeLayer` subclasses it and the PR3 `VectorQuantizeLayer` will too.
4. Replaced the `Struct` + `_coerce_proto_numbers` with a typed `FaissKmeansConfig` proto message (strictly-typed faiss kwargs).
5. Dropped the `force` param; SID sets `save_checkpoints_steps = save_checkpoints_epochs = 0`.

tiankongdeguiji (approved):

PR #541, opened 2026-06-06
[feat] Kafka: event-time driven checkpointing from message timestamp

tiankongdeguiji:
## Motivation

For online/streaming training from Kafka, `KafkaReader` previously surfaced only the message **offset** (as the resume cursor); the Kafka message **timestamp** was never read, so there was no way to checkpoint based on the *event-time* of the consumed data. This PR surfaces the message timestamp and uses it to drive checkpoint saves on consumed event-time, which is what you want for time-aligned snapshots of a stream.

## What it does

Surfaces the Kafka message timestamp on the batch and adds event-time–driven checkpoint triggers, alongside the existing step/epoch triggers:

- **`batch.data_timestamp`** — the max consumed event-time in the batch, as **Unix-epoch seconds** (`float`; `-1.0` when the source has no timestamps). `KafkaReader` normalizes its ms timestamp to seconds; `BaseDataset._build_batch` just surfaces the per-batch max (unit-agnostic, so non-Kafka datasets are unaffected).
- **New `TrainConfig` fields:**
  - `save_checkpoints_timestamp_interval` — save every N seconds of consumed event-time, aligned to the Unix epoch (wall-clock), not training epochs. `0` disables.
  - `save_checkpoints_timestamps` — absolute event-time targets (Unix-epoch seconds); save once when consumed data crosses each. Empty disables.
  - `save_checkpoints_timestamp_quorum` — fraction of workers `(0, 1]`, default `0.5`, that must have consumed past a boundary/target before a save fires.

## Design

- **Centralized save policy.** All save cadence (step / epoch / event-time) + the single per-step dedupe + the event-time watermark live in `CheckpointManager.maybe_save(...) -> bool`; the training loop calls it at the step / epoch / final points and runs eval when it returns `True`. This removes the four hand-guarded save sites in the loop (the source of an epoch-vs-timestamp double-save) and makes the decision logic unit-testable.
- **Worker quorum (not min/max).** Each rank passes its local event-time to `maybe_save`, which all-gathers across workers and reduces to the `(1 - quorum)` upper quantile — the event-time that ≥ `quorum` fraction of workers have reached. This generalizes min (`quorum=1.0`) / max (`quorum→0`), is robust to a single outlier/garbage timestamp at the default `0.5`, and counts a worker without a timestamp (`-1.0`) as "not past" so the quorum is over all workers.
- **Off the GPU.** The per-step reconciliation gathers over a dedicated CPU (gloo) process group that `CheckpointManager` creates in `set_save_policy` and owns, so it never runs on the model device or forces a per-step device sync that would stall the `TrainPipelineSparseDist` prefetch.
- **Resume-correct.** The event-time watermark is persisted in `dataloader_state` (saved every checkpoint, restored on `continue_train`) and advanced on every save, so resume re-fires no already-saved boundary and misses none. It's a single global float, identical across ranks (the per-key max-merge is a no-op on it) and ignored by the existing `dataloader_state` consumers. `dataloader_state` is typed `Dict[str, Any]` since it now holds the float watermark alongside the int offsets.
- **Distributed-safe.** The reconciliation is gated only by config (identical on all ranks) and the save decision is deterministic + identical across ranks, so the collective `save` is entered in lockstep; nothing collective runs when not saving. It relies on the same equal-step-count invariant the model's own gradient collectives already require.
- **No behavior change when unset** — with the new config fields unset, training is identical to before.

## Tests

- Pure helpers `quorum_event_time` (quantile incl. outlier/sentinel cases) and `should_save_on_timestamp` (interval/target crossings, no-refire, first-batch init) — parameterized.
- `CheckpointManager.maybe_save` — step/epoch/timestamp triggers, the dedupe (incl. the epoch-after-same-step regression), watermark stamping, restore seeding, and the `-1.0`-sentinel path; plus `_reconcile_event_time` single-process behavior (`save` mocked, no GPU/dist needed).
- `_build_batch` `data_timestamp` extraction; Kafka roundtrip asserts a positive seconds value.
- Docs: the three new `TrainConfig` fields documented in `train.md`.

### DIFF
--- a/docs/source/usage/train.md
+++ b/docs/source/usage/train.md
@@ -63,6 +63,9 @@ LR策略可以支持按epoch更新或者按step更新
 - num_epochs: 训练的epoch数
 - save_checkpoints_steps: 保存模型的步数间隔，保存模型后会做一次评估
 - save_checkpoints_epochs: 保存模型的Epoch数间隔，保存模型后会做一次评估，与save_checkpoints_steps不能同时设置
+- save_checkpoints_timestamp_interval: 按数据时间（如Kafka消息的timestamp）保存模型的间隔秒数，每当已消费数据的事件时间跨过一个按epoch对齐的间隔边界时保存一次，默认0表示关闭。仅对带timestamp的数据源（如KafkaDataset）生效
+- save_checkpoints_timestamps: 按数据时间保存模型的绝对时间点列表（单位为秒的epoch时间戳），已消费数据的事件时间每跨过一个时间点保存一次，默认为空表示关闭
+- checkpoint_timestamp_reduce: 分布式训练下各rank消费的分区不同，事件时间也不同，在触发基于时间的checkpoint前如何对各rank的事件时间做归约，可选"min"（最慢rank越过边界后才保存，默认）或"max"（任一rank越过边界即保存）
 - keep_checkpoint_max: 最多保留的最近checkpoint数量，超出后会异步删除最旧的checkpoint，默认0表示全部保留；当exporter_type为best时，会额外保留指标最优的checkpoint
 - fine_tune_checkpoint: 增量训练的checkpoint路径，也可以设置checkpoint目录，将会使用目录下最新的checkpoint
 - fine_tune_ckpt_var_map: 需要restore的参数列表文件路径，文件的每一行是{variable_name in current model}\\t{variable name in old model ckpt}
--- a/docs/source/usage/train.md
+++ b/docs/source/usage/train.md
@@ -63,9 +63,9 @@ LR策略可以支持按epoch更新或者按step更新
 - num_epochs: 训练的epoch数
 - save_checkpoints_steps: 保存模型的步数间隔，保存模型后会做一次评估
 - save_checkpoints_epochs: 保存模型的Epoch数间隔，保存模型后会做一次评估，与save_checkpoints_steps不能同时设置
-- save_checkpoints_timestamp_interval: 按数据时间（如Kafka消息的timestamp）保存模型的间隔秒数，每当已消费数据的事件时间跨过一个按epoch对齐的间隔边界时保存一次，默认0表示关闭。仅对带timestamp的数据源（如KafkaDataset）生效
-- save_checkpoints_timestamps: 按数据时间保存模型的绝对时间点列表（单位为秒的epoch时间戳），已消费数据的事件时间每跨过一个时间点保存一次，默认为空表示关闭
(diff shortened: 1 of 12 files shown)


tiankongdeguiji:
Thanks for the review — addressed in `2dd3dcc`. Summary of how each point was handled:

**Epoch-trigger double-save (real bug) — fixed at the root.** Rather than adding the missing guard to one more site, the save-cadence logic (step + epoch + event-time) is now centralized in `CheckpointManager.maybe_save(...)`, which owns `last_ckpt_step` and does a single dedupe. No call site can re-save a step another already saved. This also makes the decision logic unit-testable (it wasn't, buried in the loop) — added a regression test `test_maybe_save_dedupe_epoch_after_same_step`.

**Per-step `all_reduce` + `.item()`.** The cross-rank reconciliation is now an `all_gather` driving a **worker quorum** (replacing the `min`/`max` knob): `save_checkpoints_timestamp_quorum` (float, default 0.5) — save once ≥ quorum fraction of data-carrying workers have consumed past the boundary/target, computed as the `(1-quorum)` upper quantile. It generalizes the old knobs (`1.0`≡min, →0≡max) and is robust to a single far-future/garbage timestamp at the default, so the "untrusted timestamp" concern needs no separate wall-clock guard. (Kept per-step per design discussion; the loop does the gather, the manager does the quorum + decision with no collective when not saving.)

**"epoch-aligned" wording.** Fixed in `train.proto`, `checkpoint_util.py`, and `train.md` to clarify it means the Unix epoch (wall-clock), not training epochs.

**Test gaps.** Added: `quorum_event_time` (quantile edge cases incl. outlier robustness), `maybe_save` (step/epoch/timestamp/final + dedupe + watermark + restore seeding), `_build_batch` `data_timestamp` extraction (max valid / `-1` filter / absent→None), and `Batch.to()`/`pin_memory()` propagation (the silent-failure guard).

**"Uneven partitions → deadlock".** Not a new deadlock: a rank that drains first hangs at the model's own gradient all-reduce inside `pipeline.progress`, which precedes our collective — the equal-step-count invariant is already required by the existing collectives. Corrected the overclaiming comment to say so.

**New: resume correctness.** The event-time watermark is now persisted in `dataloader_state` and seeded on `continue_train`, so resume re-fires no already-saved boundary and misses none (advanced on every save). It's a single global float, identical across ranks, and ignored by every existing `dataloader_state` consumer.

eric-gecheng (approved):

chengaofei (approved):

PR #545, opened 2026-06-15
[feat] SID: add SidRqvae (RQ-VAE) model — STE/Gumbel-Softmax + CLIP

WhiteSwan1:
## Summary

Adds **SidRqvae**, an end-to-end differentiable RQ-VAE semantic-ID model, on top of the SID foundation already in master (`BaseSidModel` / `ResidualQuantizer` from #538, `SidRqkmeans` from #539). It complements the encoder-free, FAISS-trained `SidRqkmeans` with a gradient-trained encoder→quantizer→decoder that learns the codebook jointly with a reconstruction objective, and optionally a **pair-contrastive (masked InfoNCE)** objective.

The model runs on **both CPU and GPU** (gradient training, no FAISS dependency at train time unless `kmeans_init` warm-start is requested).

## What's in scope

**Config-driven losses.** SID losses are configured through `ModelConfig.losses`: a `LossConfig` carries one `sid_loss` `oneof` term per objective (`recon_loss`, `commitment_loss`, `contrastive_loss`), and `BaseSidModel` registers each as a `_loss_modules` entry and computes them centrally (mirroring `RankModel`). Each is a small `_Loss` module that takes its operands/mask and returns a scalar:
- `SidReconLoss` (`tzrec/loss/sid_recon_loss.py`) — per-row reconstruction distance (`recon_type` ∈ {`l2`, `l1`, `cos`}), masked-mean reduced (the mixed recon+contrastive path scores reconstruction rows only).
- `SidCommitmentLoss` (`tzrec/loss/sid_commitment_loss.py`) — VQ-VAE commitment between the encoder output and the per-layer cumulative quantized vectors (`latent_weight` `[w1, w2]`, `commitment_type`).
- `SidContrastiveLoss` (`tzrec/loss/sid_contrastive_loss.py`) — masked InfoNCE over mixed reconstruction+pair batches, driven by a per-row pair flag. Modality-agnostic (no image/text assumption), with structural masked-logit fill (dtype `finfo.min`, AMP-safe), learnable per-group temperatures (clamped before `exp`), and built-in differentiable all-gather for DDP.

**Models**
- `SidRqvae` (`tzrec/models/sid_rqvae.py`): configurable-depth encoder/decoder MLPs (mirrored; framework `MLP` + a trailing bare `Linear` so the latent/reconstruction stay unbounded), an N-layer residual vector quantizer, and an optional contrastive dual-encoder path. Knobs: `forward_mode` ∈ {`ste`, `gumbel_softmax`}, `normalize_residuals`, `distance_type`, `rotation_trick` (arXiv:2410.06424), Sinkhorn balanced assignment, and FAISS `kmeans_init` warm-start (opt-in, default off).
- Inputs flow through the framework `EmbeddingGroup` / `build_input`, so SID models consume the same grouped feature tensor as every other model and support multiple content + side-info embeddings per FORGE/PLUM. The main input is the (sole, or first-declared) feature group — auto-detected, nothing to configure; the contrastive path names its paired + pair-flag groups in `contrastive_config`.

**Modules**
- `ResidualVectorQuantizer` (`modules/sid/residual_vector_quantizer.py`): gradient-trained N-layer residual VQ. STE returns the raw codebook vector and applies one aggregate straight-through on the encoder side (the codebook trains via the commitment loss); Gumbel-Softmax's soft assignment is differentiable directly. Includes the FAISS residual warm-start (`faiss_residual_kmeans`) with a DDP fail-together broadcast that avoids the rank-0-raises-while-others-hang deadlock.
- `VectorQuantizeLayer` (`modules/sid/vector_quantize.py`): single-layer `QuantizeLayer` with STE / Gumbel-Softmax / Sinkhorn balanced assignment (Sinkhorn auto-disabled under Gumbel; `sinkhorn_epsilon` guarded `> 0` to avoid the exp-kernel overflow).
- Shared `faiss_kmeans_fit` primitive in `modules/sid/kmeans_quantize.py`, used by **both** the RQ-VAE warm-start and the RQ-KMeans offline fit (one-layer FAISS fit with an `N >= K` guard).

**Protos**
- `loss.proto`: `SidReconLoss` / `SidCommitmentLoss` / `SidContrastiveLoss` messages + the `sid_loss` `oneof` on `LossConfig`.
- `models/sid_model.proto`: `SidRqvae`, `SinkhornConfig`, and `ContrastiveConfig` messages. The main feature group is auto-detected (`group_names()[0]`), so there is no `feature_group` field.
- `model.proto`: `sid_rqvae` wired into the `ModelConfig` model `oneof`.

## Test plan

Colocated unit tests for every new/changed module (distributed paths exercised by 2-rank cases inside the same `*_test.py`), plus an end-to-end integration test:
- `sid_rqvae_test.py` — model forward/loss/metric, STE + Gumbel, with/without the contrastive path, the fail-fast group/dim guards, and the logit-scale overflow clamp.
- `sid_recon_loss_test.py`, `sid_commitment_loss_test.py`, `sid_contrastive_loss_test.py` — the three loss modules (per-row distance + masked reduction, commitment directions, masked contrastive incl. empty-mask safety, 2-rank all-gather, AMP-safe fill).
- `residual_vector_quantizer_test.py` — Gumbel grad flow vs the STE codebook gradient (trained via commitment), FAISS warm-start, non-uniform codebooks.
- `vector_quantize_test.py` — STE / Gumbel / Sinkhorn assignment.
- `residual_quantizer_test.py`, `kmeans_quantize_test.py`, `residual_kmeans_quantizer_test.py`, `quantize_layer_test.py` — the shared quantizer primitives.
- `sid_integration_test.py` — RQ-VAE and RQ-KMeans end-to-end (train → eval → predict).

All CPU tests green via `python -m tzrec.tests.run`; GPU/DDP-only cases gate on CUDA. FAISS-dependent tests skip cleanly when faiss is absent.


WhiteSwan1:
Thanks for the review — addressed in `9cf7a62`:

- **`logit_scale` overflow:** clamped to `ln(100)` before `exp()`. Good catch that `nan_to_num` only guards the CE output, not the parameter — the clamp stops the `+Inf → NaN`-grad corruption.
- **Unused CLIP outputs:** dropped `clip_acc`/`loss_self`/`ori`/`cl` and the eval-only accuracy block; `MaskedCLIPLoss` now returns just `{clip_loss}`.
- **STE double gather:** `quantize` gained an `apply_ste` flag; the RVQ caller passes `apply_ste=False` to take the raw codebook vector in a single gather (aggregate STE stays in `forward`). Bit-identical training.
- **Stale proto comment:** `RQVAE` → `SidRqvae / ResidualVectorQuantizer`.

**Test coverage** — added for every gap: Sinkhorn now has a functional balance test (vs argmin collapse); `rotation_trick` is pinned by its distinguishing property (same forward, rotated gradient ≠ plain STE); `loss_type` l1/cosine and `commitment_loss` cos branches; CLIP masked-column negatives (perturbing recon rows across all three loss groups leaves the clip-row loss intact); and the `kmeans_init` too-small-batch abort.

On the **CLIP/Sinkhorn equal-batch** note: leaving this to the framework rather than guarding in the model — the data pipeline already owns batch alignment (`rebalance=True` "equal number for each worker" + `drop_redundant_bs_eq_one` on for train/eval, and TorchRec `TrainPipelineSparseDist`'s own per-batch `all_to_all`), so the CLIP collective inherits the same guarantee. Happy to fix upstream in `calc_slice_intervals` if a real gap surfaces.

`temperature` annealing: agreed it's a real gumbel-mode limitation — will follow up separately.

tzrec/loss/clip_loss_dist_test.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) 2026, Alibaba Group;
```
tiankongdeguiji: move xxx_dist_test.py into xxx_test.py

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,357 @@
...
+
+    Args:
+        model_config (ModelConfig): an instance of ModelConfig.
+        features (list): list of features.
+        labels (list): list of label names.
+        sample_weights (list): sample weight names.
+    """
+
+    @staticmethod
+    def _build_mlp(dims: List[int]) -> nn.Sequential:
```
tiankongdeguiji: Why not use MLP module?

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,357 @@
...
+            sinkhorn_iters = cfg.sinkhorn_config.iters
+            sinkhorn_epsilon = cfg.sinkhorn_config.epsilon
+
+        self._encoder = self._build_mlp([self._input_dim, *hidden_dims, embed_dim])
+        # Decoder is the symmetric reverse of the encoder.
+        self._decoder = self._build_mlp(
+            [embed_dim, *reversed(hidden_dims), self._input_dim]
+        )
+
+        self._quantizer = ResidualVectorQuantizer(
```
tiankongdeguiji, line 97: Use `config_to_kwargs` to ensure consistency between proto parameters and module parameters, especially for sinkhorn-related kwargs.

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,357 @@
...
+            sinkhorn_iters=sinkhorn_iters,
+            sinkhorn_epsilon=sinkhorn_epsilon,
+        )
+
+        # CLIP contrastive head (optional).
+        if self._use_clip:
+            self._logit_scale_self = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
+            self._logit_scale_cl = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
+            self._logit_scale = nn.Parameter(torch.ones([]) * np.log(1 / 0.07))
+            self._masked_clip_loss_fn = MaskedCLIPLoss()
```
tiankongdeguiji: clip loss init, should be put into `init_loss` function

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def _predict_rqvae(self, embedding: torch.Tensor) -> Dict[str, torch.Tensor]:
+        """Standard RQ-VAE: encode -> quantize -> decode -> loss."""
+        result = self._forward_rqvae(embedding)
+
+        predictions: Dict[str, torch.Tensor] = {
+            "codes": result["codes"],
+        }
+
+        if self.is_train or self.is_eval:
```
tiankongdeguiji: if self.is_train or self.is_eval?

tiankongdeguiji (changes requested):

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,357 @@
...
+            losses (dict): a dict of loss tensor.
+        """
+        losses: Dict[str, torch.Tensor] = {}
+        losses["reconstruction_loss"] = predictions["reconstruction_loss"]
+        losses["quantization_loss"] = predictions["quantization_loss"]
+        if self._use_clip:
+            losses["clip_loss"] = predictions["clip_loss"]
+        return losses
+
+    def init_metric(self) -> None:
```
tiankongdeguiji: Remove it in this PR. Make metrics for SID models configurable in the next PR. `train_metrics` should be disabled by default.

tzrec/models/sid_rqvae_test.py
```diff
@@ -0,0 +1,501 @@
...
+        model_config = model_pb2.ModelConfig(
+            sid_rqvae=sid_rqvae_cfg,
+        )
+        model = SidRqvae(model_config=model_config, features=[], labels=[])
+        init_parameters(model, device=torch.device("cpu"))
+
+        for layer in model._quantizer.layers:
+            self.assertFalse(layer.use_sinkhorn)
+
+    def test_sinkhorn_config_default_enabled(self) -> None:
```
tiankongdeguiji: Use parameterized tests whenever possible, instead of expanding similar test cases manually.

tzrec/modules/sid/residual_quantizer_test.py
```diff
@@ -142,5 +149,149 @@ def test_decode_codes_sum_and_dtype(self) -> None:
         self.assertEqual(recon16.dtype, torch.bfloat16)
 
 
+class ResidualVectorQuantizerTest(unittest.TestCase):
```
tiankongdeguiji: Do not put subclass tests in the base class test file.

tzrec/modules/sid/residual_vector_quantizer_test.py
```diff
@@ -0,0 +1,163 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+
+from tzrec.modules.sid.residual_vector_quantizer import (
+    ResidualVectorQuantizer,
```
tiankongdeguiji, line 21: Add integration test into sid_integration_test.py

tzrec/modules/sid/vector_quantize.py
```diff
@@ -0,0 +1,264 @@
...
+        Args:
+            x (Tensor): input vectors, shape (B, D).
+
+        Returns:
+            Tensor: pairwise distances, shape (B, n_embed).
+        """
+        codebook = self.embedding.weight  # (n_embed, D)
+
+        if self.distance_type == "l2":
+            distances = _squared_euclidean_distance(x, codebook)
```
tiankongdeguiji: use torch.cdist(x, y, p=2).pow(2) or just cdist is ok?

tzrec/modules/sid/vector_quantize.py
```diff
@@ -0,0 +1,264 @@
...
+        # Column normalization: each sample's total weight = 1/B
+        Q /= torch.sum(Q, dim=0, keepdim=True) + 1e-8
+        Q /= B
+
+    # Step 4: scale back so columns sum to 1 (assignment)
+    Q *= B
+    return Q.t()  # (B, K)
+
+
+class VectorQuantize(QuantizeLayer):
```
tiankongdeguiji: VectorQuantize->VectorQuantizeLayer

tzrec/modules/sid/residual_vector_quantizer.py
```diff
@@ -0,0 +1,438 @@
...
+from tzrec.modules.sid.types import (
+    QuantizeForwardMode,
+    ResidualQuantizerOutput,
+)
+from tzrec.modules.sid.vector_quantize import VectorQuantize
+from tzrec.utils.logging_util import logger
+
+
+@torch.no_grad()
+def faiss_residual_kmeans(
```
tiankongdeguiji, line 32: I think `faiss_residual_kmeans` is essentially the same as `train_offline` in `ResidualKMeansQuantizer`. Should we implement `faiss_residual_kmeans` in `residual_kmeans_quantizer.py` and call it from `train_offline`?

tzrec/modules/sid/residual_vector_quantizer.py
```diff
@@ -0,0 +1,438 @@
...
+        if is_ddp:
+            dist.broadcast(enough, src=0)
+        if enough.item() == 0:
+            raise RuntimeError(
+                f"kmeans_init: rank-0 first training batch has fewer rows than the "
+                f"largest codebook ({max_k}); raise batch_size or disable kmeans_init."
+            )
+
+        if (not is_ddp) or dist.get_rank() == 0:
+            centers = faiss_residual_kmeans(
```
tiankongdeguiji, line 211: Do you consider the `kmeans_init` parameter?

tzrec/modules/sid/residual_vector_quantizer.py
```diff
@@ -0,0 +1,438 @@
...
+                data,
```
tiankongdeguiji, line 212: I think using only one batch for `faiss_residual_kmeans` is not ideal. We can optimize this in a follow-up PR.

tzrec/modules/sid/residual_vector_quantizer.py
```diff
@@ -0,0 +1,438 @@
...
+            x_unsq @ normalized_sum.unsqueeze(2) @ normalized_sum.unsqueeze(1)
+        )  # (B, 1, D)
+        rescaled_embeddings = (
+            x_unsq @ x_hat.unsqueeze(2) @ quant_hat.unsqueeze(1)
+        )  # (B, 1, D)
+        return lambda_ * (
+            x_unsq - 2 * sum_projection + 2 * rescaled_embeddings
+        ).squeeze(1)
+
+    def _train_gumbel(self) -> bool:
```
tiankongdeguiji: This is only used once, so it probably doesn’t need to be a separate function.

tiankongdeguiji (changes requested):

tzrec/modules/sid/vector_quantize.py
```diff
@@ -0,0 +1,264 @@
...
+        """Assign ``x`` to the codebook (the :class:`QuantizeLayer` interface).
+
+        Commitment loss is computed by the caller
+        (:meth:`ResidualVectorQuantizer._single_commitment_loss`); device follows
+        ``x``, so this runs on CPU or GPU unchanged.
+
+        Args:
+            x (Tensor): input vectors, shape (B, D).
+            temperature (float): temperature for Gumbel-Softmax.
+            apply_ste (bool): in STE training, wrap the embedding with the
```
tiankongdeguiji: Why do we need an `apply_ste` parameter? I think this can be derived from `self.forward_mode == QuantizeForwardMode.STE`.

tzrec/modules/sid/vector_quantize.py
```diff
@@ -0,0 +1,264 @@
...
+    ) -> QuantizeOutput:
+        """Assign ``x`` to the codebook (the :class:`QuantizeLayer` interface).
+
+        Commitment loss is computed by the caller
+        (:meth:`ResidualVectorQuantizer._single_commitment_loss`); device follows
+        ``x``, so this runs on CPU or GPU unchanged.
+
+        Args:
+            x (Tensor): input vectors, shape (B, D).
+            temperature (float): temperature for Gumbel-Softmax.
```
tiankongdeguiji: Why do we need a separate `temperature` parameter? I think `temperature` should be a class initialization parameter.

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,358 @@
...
+        z_e1 = self._encode(fea1)
+        quant1 = self._quantizer(z_e1, temperature=temperature)
+        x_hat1 = self._decode(quant1.quantized_embeddings)
+
+        z_e2 = self._encode(fea2)
+        quant2 = self._quantizer(z_e2, temperature=temperature)
+        x_hat2 = self._decode(quant2.quantized_embeddings)
+
+        recon_mask = ~clip_mask
+        recon_loss = self._recon_loss(x_hat1, fea1, recon_mask)
```
tiankongdeguiji: I think the loss computation should be refactored into `def loss()`.

tzrec/modules/sid/residual_vector_quantizer.py
```diff
@@ -0,0 +1,438 @@
...
+    Args:
+        embed_dim (int): dimension of input embeddings.
+        n_layers (int): number of quantization layers.
+        n_embed (int|List[int]): codebook size per layer. Default: 256.
+        forward_mode (str): VQ forward mode ('ste'|'gumbel_softmax').
+            Default: 'ste'.
+        normalize_residuals (bool): L2-normalize residuals before each
+            quantization layer. Default: False.
+        distance_type (str): distance metric, 'l2' or 'cosine'. Default: 'l2'.
+        commitment_loss (str): commitment loss type, 'l2', 'l1' or 'cos'.
```
tiankongdeguiji: I think we should add a separate commitment loss class and use it in `sid_rqvae`, rather than implementing it here.

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,358 @@
...
+        model_config: ModelConfig,
+        features: List[BaseFeature],
+        labels: List[str],
+        sample_weights: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(model_config, features, labels, sample_weights, **kwargs)
+
+        cfg = self._model_config  # SidRqvae proto message
+        self._loss_type = cfg.loss_type
```
tiankongdeguiji: I think it would be better to add a `oneof sid_loss` field to `LossConfig`, covering losses such as `recon_l2_loss`, `recon_l1_loss`, `clip_loss`, `commitment_loss`, etc. Then we can implement `init_loss` and centralize the loss computation logic in `sid_model.py`, similar to `rank_model.py`.

tiankongdeguiji (changes requested):

tzrec/loss/clip_loss.py
```diff
@@ -0,0 +1,187 @@
...
+from typing import Dict, List, Optional
+
+import torch
+import torch.distributed as dist
+import torch.distributed.nn as dist_nn
+from torch.nn import functional as F
+from torch.nn.modules.loss import _Loss
+
+
+class MaskedCLIPLoss(_Loss):
```
tiankongdeguiji: I don’t think this should be named `clip`. It is an InfoNCE loss, so the naming should not be tied to concepts like `image_embed` and `text_embed`. Also, I think it would be better to gather valid embedding pairs outside the loss function, instead of performing all-gather and matmul over all embeddings.

tzrec/models/sid_rqvae.py
```diff
@@ -0,0 +1,331 @@
...
+    def predict(self, batch: Batch) -> Dict[str, torch.Tensor]:
+        """Predict the model.
+
+        Args:
+            batch (Batch): input batch data.
+
+        Return:
+            predictions (dict): a dict of predicted result.
+        """
+        embedding = self._extract_feature(batch)
```
tiankongdeguiji: I think it would be better to use `build_input` with feature groups instead of `_extract_feature`. FORGE shows that incorporating sideinfo works better, and PLUM also uses multiple content embeddings as inputs.

tiankongdeguiji (changes requested):

WhiteSwan1:
Thanks for the careful reviews @tiankongdeguiji — here's a consolidated summary of the changes. The smaller fixes are listed first, then the larger refactors and the one suggestion I'd handle a bit differently. All of these changes have been validated through end-to-end experiments and have no impact on model results.

## Smaller fixes (done)

- Merged the `*_dist_test.py` files into the colocated `*_test.py` and deleted the separate dist files.
- Used `config_to_kwargs` for the sinkhorn kwargs so the proto and module parameters stay in sync.
- Dropped the incorrect `is_train or is_eval` phase check.
- Removed the train-time metrics for now; I'll make SID metrics configurable in a follow-up, disabled by default.
- Switched the manually-expanded test cases to `parameterized`.
- Moved the subclass tests out of the base-class test file.
- Added an end-to-end integration test (RQ-VAE and RQ-KMeans) in `sid_integration_test.py`.
- Renamed `VectorQuantize` → `VectorQuantizeLayer`.
- Inlined the single-use `_train_gumbel` helper.
- Made `temperature` a class-init parameter instead of a `quantize()` argument.
- Switched the encoder/decoder to the framework `MLP` (plus a bare trailing `Linear` for the unbounded latent/reconstruction projection, since `MLP` always activates its last layer); the rewrite is behavior-preserving.
- Replaced the hand-rolled squared-L2 distance with `torch.cdist(x, y, p=2).pow(2)`. I checked it matches the previous form to fp32 noise, including a coincident zero-distance point (finite gradient, no NaN).

## Config-driven losses

I added a `oneof sid_loss` to `LossConfig` (reconstruction / commitment / contrastive), merged the three reconstruction variants into a single one with a `recon_type` field, and centralized the loss handling in `BaseSidModel` (`init_loss()` + `loss()`), the way `rank_model` does. `predict()` now returns only the raw tensors and `loss()` computes the configured terms. I also extracted a standalone commitment-loss class and use it from the model, and moved the contrastive loss and its learnable temperatures into `init_loss()`.

## Input via `build_input` + feature groups

The SID models now consume the framework's `build_input` / `EmbeddingGroup` path like every other model instead of reading the embedding out of the batch directly, and `input_dim` is derived from the feature group's total dimension. This is the change you suggested for side-info / multiple content embeddings (FORGE/PLUM): the main feature group can now hold one content embedding or several content + side-info features concatenated. The paired embedding and the pair flag for the contrastive path are feature groups too. I updated the proto, the mock configs, and the unit tests accordingly.

## Contrastive loss naming

Renamed `MaskedCLIPLoss` → `MaskedInfoNCELoss` and made it modality-agnostic (no `image_embed` / `text_embed`); the model now owns the pairing structure and passes the loss a generic pair mask.

## On gathering the valid pairs outside the loss — keeping the masking inside now

Three reasons:

1. **It's numerically identical.** Gather-then-mask produces the exact same result as filter-then-gather in the forward: non-pair columns are masked with the dtype's most-negative finite value (so they drop out of the softmax, `exp(...) = 0`), and the pair rows are normalized by the valid-pair count. So there's no accuracy difference to gain.

2. **The negatives are cross-rank.** The loss builds its negatives by `all_gather`-ing the embeddings across ranks and scoring each local row against the global bank. If I filter the pairs *before* the gather, each rank gathers a different (ragged) row count — a size mismatch in the collective — and on a step where a rank has no pairs, `logit_scale` drops out of that rank's autograd graph, which hangs DDP's gradient sync. Keeping the full batch through the gather avoids both.

3. **It keeps the loss `torch.compile`-friendly.** The loss is deliberately written branch-free with fixed `(B, B_global)` shapes. Filtering by the pair mask makes the row count — and every downstream tensor shape — data-dependent, which triggers guard failures / recompiles (or graph breaks) under `torch.compile`; the masked form keeps the shapes static so the graph traces once.

Based on the above considerations, MaskedInfoNCELoss‘s masking did not make any changes in this PR. If necessary, this module will be carefully evaluated separately in the future PR to ensure that the filter-then-compute runs correctly in cross-rank scene.

## To be completed in a follow-up PR

- Unifying the single-batch FAISS warm-start with the reservoir `train_offline` path: they currently sit in different classes with different responsibilities, so merging them is a larger refactor of where the FAISS fit lives — I'll handle it in a follow-up rather than expand this PR's scope.
- Replacing the single-batch FAISS fit with a proper multi-batch fit (as we discussed); a `# TODO` marks the spot.
- `kmeans_init` is kept as an opt-in for now, with documented caveats (it seeds from one batch, and gradient training refines it afterward).

tzrec/protos/models/sid_model.proto
```diff
@@ -37,6 +113,8 @@ message SidRqkmeans {
     // are wasted.
     optional uint32 train_sample_size = 6 [default = 0];
 
-    // Name of the item embedding feature inside the input Batch.
-    optional string embedding_feature_name = 40 [default = "item_emb"];
+    // Name of the main input FEATURE GROUP (built by the model's EmbeddingGroup
+    // from ModelConfig.feature_groups). May hold one or many content/side-info
+    // features; their concatenated dim is the K-Means dimension.
+    optional string feature_group = 40 [default = "deep"];
```
tiankongdeguiji: not needed, auto detect group_names()[0]

tiankongdeguiji (changes requested):

tiankongdeguiji (approved):

PR #546, opened 2026-06-16
[feat] add model delta tracker

eric-gecheng:


tzrec/utils/delta_embedding_dump.py
```diff
@@ -0,0 +1,766 @@
...
+        if isinstance(local_value, torch.Tensor):
+            info = _table_shard_info_from_tensor(local_value, shard_info)
+            return _TableWeight(tensor=local_value, shard_info=info)
+    if isinstance(value, torch.Tensor):
+        info = _table_shard_info_from_tensor(value, shard_info)
+        return _TableWeight(tensor=value, shard_info=info)
+    raise TypeError(f"Unsupported embedding table value type: {type(value)}")
+
+
+class DeltaEmbeddingDumper:
```
tiankongdeguiji, line 309: Please add argument docstrings for public classes and public functions.

tiankongdeguiji (changes requested):

tzrec/main.py
```diff
@@ -589,6 +615,14 @@ def train_and_evaluate(
     is_rank_zero = int(os.environ.get("RANK", 0)) == 0
     is_local_rank_zero = int(os.environ.get("LOCAL_RANK", 0)) == 0
     acc_utils.allow_tf32(train_config)
+    if train_config.HasField("delta_embedding_dump_config"):
```
tiankongdeguiji: I think it would be cleaner to keep the validation inside `DeltaEmbeddingDumper`. We can check whether the module contains a `ShardedManagedCollisionCollection`.

tiankongdeguiji (approved):

PR #557, opened 2026-06-30
[feat] bump graphlearn to 1.3.8, drop numpy<2 pin

tiankongdeguiji:
## What

- Bump `graphlearn` 1.3.7 → 1.3.8 (new wheels under `third_party/graphlearn/`).
- Drop the long-standing `numpy<2` pin.

## Why

graphlearn 1.3.7 was the only dependency forcing `numpy<2`. graphlearn 1.3.8
adds numpy>=2 support, so the pin is no longer needed.

## Verification

Staged so CI proves numpy 2.x compatibility before the pin is relaxed:

1. First push pins `numpy>2` to force the numpy 2.x path through the full CI
   matrix (GPU / CPU / build / codestyle / pytyping).
2. Once green, the pin is relaxed to unconstrained `numpy`.

### DIFF
--- a/.github/workflows/buildtest_ci.yml
+++ b/.github/workflows/buildtest_ci.yml
@@ -22,6 +22,6 @@ jobs:
         run: |
           cd run_${{ github.run_id }}
           bash scripts/build_wheel.sh nightly
-          find dist/ -name '*.whl' | xargs pip install
+          find dist/ -name '*.whl' | xargs -I{} pip install "{}[gpu]"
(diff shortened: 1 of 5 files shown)


tiankongdeguiji:
Superseded by #551 (the 1.3.0 bump), which already includes these changes: numpy 2, faiss upgraded to a numpy-2 build (1.14.3+cu12), graphlearn 1.3.8, `sampler.py` `np.string_`→`np.bytes_`, and the build-test install fix. Closing as redundant.

PR #593, opened 2026-07-08
[ci] code review: add Codex review job triggered by codex-review label

tiankongdeguiji:
## Summary

Adds an OpenAI Codex PR-review path to `code_review.yml`, alongside the existing Claude flow: a collaborator adds the `codex-review` label and the same self-hosted runner runs a headless `codex exec` review, posting inline comments through the existing bun MCP server and a summary via `gh pr comment`.

**Single source of truth for review content.** No second set of review markdowns: `scripts/ci/gen_codex_review.py` converts the trusted base-branch `.claude/agents/*.md` into Codex subagent TOMLs (`.codex/agents/<name>.toml`, `sandbox_mode = "read-only"` mirroring the Claude agents' read-only tools) and turns `.claude/commands/review-pr.md` into the exec prompt at runtime. The one shared wording change makes the inline-comment instruction tool-neutral so both CLIs resolve their own MCP tool naming.

**Design notes**
- Same trusted-file pattern as the Claude job (GHSA-f9x3-9rgg-92p7): `.claude/` + converter are checked out from `base.sha`; the PR's `.claude/`, `.codex/`, and `AGENTS.md` are dropped, and trusted `CLAUDE.md` is installed as `AGENTS.md` (Codex auto-reads it — supplies project context and blocks a PR-authored injection file).
- Sandbox is `workspace-write` with `sandbox_workspace_write.network_access=true` (gh needs network), never `danger-full-access`, so runner home / stored credentials stay unreadable to the model.
- MCP server config is passed per-run via `-c` overrides; `env_vars` forwards `GITHUB_TOKEN`/`REPO_OWNER`/`REPO_NAME`/`PR_NUMBER` to the server by name only — no secret values in argv or on-disk config.
- Auth is stored ChatGPT OAuth from a one-time `codex login --device-auth` on the runner (mirrors `claude /login`); no `OPENAI_API_KEY` secret.
- Concurrency groups now include the label name, so codex and claude reviews of the same PR no longer cancel each other (previously any label event shared one group).

## Known limitations / follow-ups
- `pull_request_target` runs the base-branch workflow definition, so this job can only be exercised after merge: do the one-time runner setup in the yml header (install Codex CLI, `codex login --device-auth`), then add the `codex-review` label to a test PR. Requires a recent Codex CLI (subagents + `mcp_servers.*.env_vars`).
- The `codex-review` label must exist in the repo before first use.

## Verification
- Converter run against the repo's real `.claude/`: 5 agent TOMLs generated, all parse with `tomllib`, prompt assembles with REPO/PR header.
- Workflow YAML parses; both jobs and label-scoped concurrency verified.
- `pre-commit` passes on all changed files.

https://claude.ai/code/session_012Lbq87BJ7AiDFqs7srcP6v

### DIFF
--- a/.claude/commands/review-pr.md
+++ b/.claude/commands/review-pr.md
@@ -14,7 +14,7 @@ Perform a comprehensive code review using subagents for key areas:
 Instruct each to only provide noteworthy feedback. Once they finish, review
 the feedback and post only the feedback that you also deem noteworthy.
 
-Use `mcp__github_inline_comment__create_inline_comment` for inline comments on specific lines.
+Use the `create_inline_comment` tool of the `github_inline_comment` MCP server for inline comments on specific lines.
 Use top-level comments for general observations or praise.
 Use `gh pr comment` for top-level summary comments.
 Do NOT use `gh api` — it is not available.
--- a/.claude/commands/review-pr.md
+++ b/.claude/commands/review-pr.md
@@ -3,6 +3,9 @@ allowed-tools: Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*)
 description: Review a pull request
 ---
 
+Fetch the PR context with `gh pr view` and the changes with `gh pr diff`.
+Review statically — do not run tests or builds.
+
 Perform a comprehensive code review using subagents for key areas:
 
 - code-quality-reviewer
--- a/.github/workflows/code_review.yml
+++ b/.github/workflows/code_review.yml
@@ -191,8 +191,7 @@ jobs:
             --out-dir "${{ runner.temp }}/codex-gen" \
             --repo ${{ github.repository }} \
             --pr-number ${{ github.event.pull_request.number }} \
-            --prompt-out "${{ runner.temp }}/codex-prompt.md" \
-            --agents-md trusted-claude/AGENTS.md
+            --prompt-out "${{ runner.temp }}/codex-prompt.md"
           bash trusted-claude/scripts/ci/use_trusted_agent_files.sh \
             trusted-claude pr-code "${{ runner.temp }}/codex-gen"
 
@@ -226,12 +225,13 @@ jobs:
 
           # Sandbox has no network; only the gh pr view/diff/comment rules allowlist
           # escapes it (claude-job parity). Unstripped TOKEN env so gh can auth.
+          # Every AGENTS.md codex can see is trusted: the workspace one is restored by
+          # use_trusted_agent_files.sh and the per-run CODEX_HOME has none.
           codex exec \
             --ignore-user-config \
             --ephemeral \
             --sandbox workspace-write \
             -c shell_environment_policy.ignore_default_excludes=true \
-            -c project_doc_max_bytes=0 \
             -c model_reasoning_effort='"medium"' \
             -c mcp_servers.github_inline_comment.command="\"${BUN_PATH}\"" \
             -c 'mcp_servers.github_inline_comment.args=["run","/opt/claude-code-action/src/mcp/github-inline-comment-server.ts"]' \
--- a/.github/workflows/code_review.yml
+++ b/.github/workflows/code_review.yml
@@ -58,7 +58,8 @@
 #   codex login --device-auth
 #   codex login status
 #
-#   # Optionally pin the model in ~/.codex/config.toml, e.g. model = "gpt-5-codex"
+#   # The job runs with --ignore-user-config (~/.codex/config.toml is not loaded);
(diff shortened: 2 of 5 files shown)


tiankongdeguiji:
Follow-up bd6def93 after a full end-to-end run of the codex-review job against a real fork PR (the first live run failed at the trusted-file step): the setup step no longer references untracked `CLAUDE.md` (AGENTS.md injection is instead blocked wholesale via `project_doc_max_bytes=0`), and `codex exec` is hardened for headless CI — MCP tools pre-approved (exec mode otherwise auto-cancels them), `GITHUB_TOKEN` visible to `gh` (codex strips `*TOKEN*` env vars from shell commands by default), runner user's personal config/skills isolated via `--ignore-user-config` + HOME shim, reasoning effort pinned to medium, and the trusted-file swap committed in the workspace so the review sees only the PR's own diff.

Verified end-to-end on the review runner: 5 subagents spawned from the generated `.codex/agents/`, inline comment posted via the MCP server, and a summary comment via `gh pr comment`.

tiankongdeguiji:
Review triage (65a4a19a):

**Fixed**
- `cp CLAUDE.md` blocker — already removed in bd6def93 (it was the live CI failure); AGENTS.md injection is instead disabled wholesale via `project_doc_max_bytes=0`.
- `encoding="utf-8"` on all converter file I/O and `mkdir(exist_ok=True)` — both applied; converter re-verified under `LC_ALL=C` with a double-run and `tomllib` round-trip of all generated agent TOMLs.

**Addressed differently**
- Nested-`AGENTS.md` sweep: `project_doc_max_bytes=0` (bd6def93) turns off the AGENTS.md mechanism entirely — root, nested, and user-global — so the tree-wide `find`-delete would be a redundant second layer. The remaining kernel (workspace-write confines writes, not reads, while `gh` needs network) can't be closed by config today; it is now documented in the header as part of the label-gated trust model, with guidance to keep the runner single-purpose and its credentials minimal.

**Not taken**
- Composite action / emitting `config.toml`: per-run `-c` overrides were chosen deliberately (no on-disk runner config to drift, secrets forwarded by name only); the two jobs share shape but differ enough that the indirection outweighs the dedup.
- Converter unit test: `tzrec/tests/run.py` only discovers tests under `tzrec/`, so a `scripts/ci` test would never execute in CI; the tomllib round-trip is performed as part of change verification instead.
- Unclosed front-matter fence / `sandbox_mode` hardcoding: converter inputs are trusted repo files, and read-only is the intended policy for review agents regardless of their Claude-side `tools` — a loud failure at CI setup is acceptable for malformed input.

tiankongdeguiji:
Review triage, round 2 (cbcb275a..c6c2fda8) — all three findings were true issues:

**No-op `git commit` aborts claude-review (should-fix)** — confirmed and fixed in cbcb275a. Reproduced exactly as described (clean clone → swap is a no-op → `git commit` exit 1 under `set -e`); the staged-diff guard now skips the commit, and a real instruction-file diff still produces the setup commit.

**Token exposure via `ignore_default_excludes=true` (consider)** — the objective was right but the mechanism claim doesn't hold at runtime: `shell_environment_policy` is inert in codex 0.142.4. Verified live that model-run commands see the token with the flag absent, with an explicit `exclude=["GITHUB_TOKEN","GH_TOKEN"]`, and with `inherit="core"` — so there was no built-in stripping to preserve, and the flag being removed was already a no-op. c6c2fda8 fixes it at the source instead: codex runs under `env -u GITHUB_TOKEN -u GH_TOKEN`, `gh` authenticates from a stored login in the shim HOME (out-of-band `gh auth login --with-token`, as suggested), and the MCP server gets its env from a generated `config.toml` in the per-run CODEX_HOME — which also replaces the `-c` override forest and `--ignore-user-config`. Result is tighter than proposed: `printenv` in model commands returns unset (verified live, along with `gh pr view` via stored login and an inline comment through the config-file MCP wiring). Remaining surface is file-reads (`hosts.yml`, per-run `config.toml`) feeding the accepted `gh pr comment --body` sink.

**json.dumps non-BMP surrogate escapes (consider)** — true; fixed properly in 1bad992c with `ensure_ascii=False` (raw UTF-8 is valid in TOML basic strings), verified by round-tripping an emoji/CJK agent file through `tomllib`, rather than documenting the ASCII caveat.

One CI-only residual to watch on first post-merge run: `gh auth login --with-token` with the Actions installation token is exercised only in CI (local verification used a user token); if it rejects the token the job fails loudly at the trusted step.

WhiteSwan1 (approved):

eric-gecheng (approved):

PR #601, opened 2026-07-15
[bugfix] avoid tensor-sized temporaries when trunc_normal_ inits large embedding tables

tiankongdeguiji:
## Problem

`nn.init.trunc_normal_` in torch>=2.12 was rewritten with rejection sampling (pytorch/pytorch#174997). The new implementation materializes several temporaries as large as the tensor itself: the bounds check `mask = (result < lo) | (result > hi)` alone allocates up to three tensor-sized bool masks, and any resample iteration adds two fp32 buffers of full tensor size.

torchrec applies a table's `init_fn` to the full local shard at DMP sharding time (`ShardedEmbeddingBagCollection._initialize_torch_state -> reset_parameters`). For a very large table this transient allocation is huge — e.g. a 150M-row x dim-64 ZCH table with a `fused_uvm_caching` kernel needs an 8.94 GiB first mask (and ~27 GiB peak at that line) — and it OOMs GPUs that ran the exact same config fine on torch<=2.11, where `trunc_normal_` was fully in-place (`uniform_ -> erfinv_ -> mul_ -> add_ -> clamp_`, ~0 extra bytes).

## Fix

- Add `tzrec/utils/init_util.py`:
  - `create_init_fn(expr)` centralizes the eval of config `init_fn` expressions (previously inline `eval(f"partial({expr})")` at three sites) with an explicit `{partial, nn, torch}` namespace, and swaps `nn.init.trunc_normal_` for an env-switchable `trunc_normal_`.
  - `trunc_normal_` defers to native `nn.init.trunc_normal_` by default. Setting `USE_INPLACE_TRUNC_NORMAL=1` initializes with the in-place inverse-CDF implementation of torch<=2.11 (~0 transient HBM), restoring for memory-constrained huge-table jobs exactly the initialization their configs trained with on torch<=2.11.
- The `zch.threshold_filtering_func` eval in `feature.py` previously resolved `partial`/`nn` from module globals kept alive by `# NOQA` imports; it now gets an explicit namespace (`partial`, `nn`, `torch`, and the three `*_threshold_filter` helpers) instead of leaking module globals into config eval.
- Bump version to 1.3.4.

An earlier revision dispatched on a `numel > 2**27` threshold; review correctly noted that a fixed element cutoff neither bounds native temporaries just below it (~1.4 GiB possible) nor avoids silently reintroducing the torch<=2.11 endpoint-clamp tail defect above it (a `uniform_` sample landing exactly on an interval endpoint clamps to a/b after `erfinv_`; measured ~2^-24 per element, ~450 exact -2.0 outliers on a 150M x 64 table). The explicit opt-in keeps native statistics for everyone by default and makes the memory/statistics trade a deliberate per-job choice.

## Test Plan

- `tzrec/utils/init_util_test.py` (7 tests): expression parsing and `trunc_normal_` swap (kwargs preserved, non-trunc-normal init fns untouched), default path calls native `nn.init.trunc_normal_` with exact argument forwarding, `USE_INPLACE_TRUNC_NORMAL=1` path calls the in-place implementation with exact argument forwarding, statistics through the public API with nonzero mean (mean=0.5, std=0.0125), truncation at ±1σ, and generator reproducibility.
- `tzrec/features/id_feature_test.py`: added a `partial(dynamic_threshold_filter, ...)` case for `zch.threshold_filtering_func` to guard the explicit eval namespace.
- Ran `tzrec.utils.init_util_test` (7), `tzrec.features.id_feature_test` (58), `tzrec.modules.embedding_test` (52), `tzrec.models.wide_and_deep_test` (3, covers `wide_init_fn`) — all pass; `pre-commit` clean.
- End-to-end: built an `IdFeature` with `init_fn: "nn.init.trunc_normal_,mean=0.0,std=0.0125"` and verified the config-created init_fn calls native torch by default and the in-place path with `USE_INPLACE_TRUNC_NORMAL=1`, with correct statistics on both.

https://claude.ai/code/session_01Lu52Dws7J8GLgr7Hzviisn

### DIFF
--- a/tzrec/features/feature.py
+++ b/tzrec/features/feature.py
@@ -14,14 +14,14 @@ import os
 import shutil
 from collections import OrderedDict
 from copy import copy
-from functools import partial  # NOQA
+from functools import partial
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pyarrow as pa
 import pyfg
 import torch
-from torch import nn  # NOQA
+from torch import nn
 from torchrec.distributed.planner.types import ParameterConstraints
 from torchrec.modules.embedding_configs import (
     BaseEmbeddingConfig,
@@ -35,9 +35,9 @@ from torchrec.modules.mc_modules import (
     LRU_EvictionPolicy,
     ManagedCollisionModule,
     MCHManagedCollisionModule,
-    average_threshold_filter,  # NOQA
-    dynamic_threshold_filter,  # NOQA
-    probabilistic_threshold_filter,  # NOQA
+    average_threshold_filter,
+    dynamic_threshold_filter,
+    probabilistic_threshold_filter,
 )
 from torchrec.types import DataType
 
@@ -60,7 +60,7 @@ from tzrec.modules.dense_embedding_collection import (
 from tzrec.protos import feature_pb2
 from tzrec.protos.data_pb2 import FgMode
 from tzrec.protos.feature_pb2 import FeatureConfig, SequenceFeature
-from tzrec.utils import config_util, dynamicemb_util, env_util
+from tzrec.utils import config_util, dynamicemb_util, env_util, init_util
 from tzrec.utils.load_class import get_register_class_meta
 from tzrec.utils.logging_util import logger
 
@@ -615,7 +615,7 @@ class BaseFeature(object, metaclass=_meta_cls):
             embedding_name = self.config.embedding_name or f"{self.name}_emb"
             init_fn = None
             if self.config.HasField("init_fn"):
-                init_fn = eval(f"partial({self.config.init_fn})")
+                init_fn = init_util.create_init_fn(self.config.init_fn)
             emb_bag_config = EmbeddingBagConfig(
                 num_embeddings=self.num_embeddings,
                 embedding_dim=self._embedding_dim,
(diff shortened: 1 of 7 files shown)


tiankongdeguiji:
Addressed the review findings in daf9e27e:

1. **Endpoint-clamp outliers of the inverse-CDF path** (init_util.py:57) — confirmed empirically: on torch 2.11 (identical algorithm), a 2^30-element fp32 CUDA fill with std=0.0125 produces 51 elements exactly at -2.0 (~2^-24/element, single-sided since `uniform_` includes `from`), i.e. ~450 outliers on a 150M×64 table. Resolved together with the threshold concern below: the in-place path is no longer the default for any size. `trunc_normal_` now defers to native `nn.init.trunc_normal_` unless `USE_INPLACE_TRUNC_NORMAL=1` is set, which opts a memory-constrained job into exactly the torch<=2.11 initialization behavior (defect included, as documented in the docstring) that such configs were trained with for years.

2. **`2**27` cutoff can still OOM below the boundary / ignores dtype** (init_util.py:92) — agreed; the fixed element cutoff was a heuristic, not a bound. The threshold is removed entirely in favor of the explicit env switch, so the memory/statistics trade is a deliberate per-job choice rather than a silent size-dependent behavior change.

3. **`add_(mean)` untested with nonzero mean** (init_util_test.py:50) — fixed; the statistics test now fills with mean=0.5, std=0.0125 through the public API and asserts the mean.

4. **std=0 / a>=b degenerate configs "silently fill with NaNs"** (init_util.py:55) — not reproducible; the failure is already loud on both paths: `std=0.0` raises `ZeroDivisionError: float division by zero` (Python float division in `norm_cdf`), and `a > b` raises `RuntimeError: uniform_ expects to return a [from, to) range`. `a == b` fills the constant `a`, identical to torch<=2.11 native behavior. No guard added.

chengaofei (approved):

PR #603, opened 2026-07-16
[feat] delta embedding dump upload to PAI Feature Store

eric-gecheng:


tzrec/tools/feature_store/check_feature_store_delta.py
```diff
@@ -0,0 +1,551 @@
...
+    output_dir: str,
+    file_prefix: str,
+    global_step: int,
+    expected_shards: int,
+) -> List[str]:
+    """Resolve the canonical parquet shards for one committed upload step."""
+    single_rank_path = os.path.join(
+        output_dir, f"{file_prefix}_step_{global_step}.parquet"
+    )
+    if os.path.isfile(single_rank_path):
```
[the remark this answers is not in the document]
eric-gecheng: this file is only for offline check purpose.

requirements/feature_store.txt
```diff
@@ -0,0 +1 @@
+feature_store_py @ https://gecheng-rec.oss-accelerate-overseas.aliyuncs.com/software/feature_store_py-2.2.7-py3-none-any.whl
```
tiankongdeguiji: Please move feature_store_py into requirements/runtime.txt and resolve any dependency incompatibilities at the runtime dependency level. Also use the officially maintained package url under feature-store-py oss?

tzrec/protos/train.proto
```diff
@@ -29,16 +29,76 @@ message GradClipping {
     optional bool enable_global_grad_clip = 4 [default = false];
 }
 
+message FeatureStoreConfig {
+    // Runtime credentials are read only from ALIBABA_CLOUD_ACCESS_KEY_ID,
+    // ALIBABA_CLOUD_ACCESS_KEY_SECRET, FEATUREDB_USERNAME and FEATUREDB_PASSWORD.
+    reserved 2 to 5;
```
tiankongdeguiji: why 
reserved 2 to 5
reserved "access_key_id", "access_key_secret", "featuredb_username", "featuredb_password" ?

tzrec/main.py
```diff
@@ -452,11 +467,22 @@ def run_eval(step: int, epoch: int) -> None:
...
+    )
     for i_epoch in epoch_iter:
+        # Multi-rank dumps synchronize their per-step state, so uneven workers
+        # must fail together instead of silently dropping a lagging rank's
+        # trailing delta rows or leaving collectives.
         pipeline = create_train_pipeline(
             model,
             optimizer,
-            check_all_workers_data_status=check_all_workers_data_status,
+            check_all_workers_data_status=sync_train_data_exhaustion,
```
tiankongdeguiji: I don’t think `require_equal_train_batches` or `fail_on_uneven_data` is necessary. Regular distributed dataloaders already produce equal batch counts, while `batch_cost_size` enables `check_all_workers_data_status`, which stops all ranks at the shortest iterator before the extra batch is processed. This change adds an unnecessary per-batch collective and turns expected `batch_cost_size` imbalance into a fatal error.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+    """A multi-rank canonical shard set is still being atomically replaced."""
+
+
+@dataclass(frozen=True)
+class FeatureStoreUploadSettings:
+    """Validated immutable settings copied from the runtime protobuf."""
+
+    region: str
+    endpoint: str
+    access_key_id: str = field(repr=False)
```
tiankongdeguiji: Please follow the credential-provider pattern already used in `tzrec/datasets/odps_dataset.py` instead of representing cloud credentials in `FeatureStoreConfig`. In particular, `security_token` is a bearer credential and should not be a protobuf/config field. AK/SK/STS credentials should be obtained at runtime through the Alibaba Cloud credential provider (`CredentialProviderAccount` or its underlying credentials client). If `FeatureStoreClient` only accepts scalar credential arguments, adapt the provider when creating or recreating the client so refreshed STS credentials can be picked up.

Once `FeatureStoreConfig` contains no credentials, the `config_util` sanitization changes are unnecessary. Please remove `security_token` from the proto and revert `save_pipeline_config_artifact` and its call-site changes, keeping the existing `save_message` path. Configuration artifacts should remain semantically complete and replayable; secrets should never enter the config in the first place.

tiankongdeguiji (changes requested):

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+    def close(self, raise_on_error: bool = True, drain: bool = True) -> None:
+        """Close the rank-zero uploader; abnormal shutdown can skip draining."""
+        if self._uploader is not None:
+            self._uploader.close(raise_on_error=raise_on_error, drain=drain)
+
+    def _feature_store_upload_error(
+        self, force: bool = False
+    ) -> Optional[BaseException]:
+        """Collect a local/shared uploader error without changing rank control flow."""
+        if not getattr(self, "_feature_store_enabled", False):
```
tiankongdeguiji: These attributes are initialized unconditionally in __init__, so the getattr(..., default) fallbacks hide broken object invariants—for example, a missed assignment could silently disable uploader error propagation. Please access the fields directly.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+            self._feature_store_enabled,
         )
 
     def clear(self) -> None:
         """Clear tracked sparse ids, usually after restore-time dummy steps."""
-        self._tracker.clear(_CONSUMER)
+        self._tracker.clear()
+
+    @property
+    def requires_synced_dataloader_exhaustion(self) -> bool:
```
tiankongdeguiji: do not need

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+            self._uploader.start()
+        if self._interval_secs is not None:
+            self._next_dump_time = time.monotonic() + self._interval_secs
+
+    def close(self, raise_on_error: bool = True, drain: bool = True) -> None:
+        """Close the rank-zero uploader; abnormal shutdown can skip draining."""
+        if self._uploader is not None:
+            self._uploader.close(raise_on_error=raise_on_error, drain=drain)
+
+    def _feature_store_upload_error(
```
tiankongdeguiji: `torchrun` terminates all workers when any rank exits with an uncaught exception, so this function does not need to return an error for later `all_reduce` propagation. Please raise the error directly and remove the error-bit rendezvous.

tiankongdeguiji (changes requested):

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+            raise RuntimeError("FeatureStore delta dumper must be started before use")
+        value = f"{self._run_generation}:{global_step}".encode("ascii")
+        return hashlib.sha256(value).hexdigest()[:32]
+
+    def _build_sparse_embedding_contract(
+        self,
+    ) -> Tuple[Dict[str, SparseEmbeddingIdentity], Dict[str, int]]:
+        """Build the same physical-table identity consumed by sparse export."""
+        metadata_by_identity: Dict[Tuple[str, str], Tuple[int, Tuple[str, ...]]] = {}
+        owner_by_identity: Dict[Tuple[str, str], str] = {}
```
tiankongdeguiji: owner -> fqn for readability

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+            for table_name, table_config in table_name_to_config.items():
+                dimension = _int_attr(table_config, "embedding_dim")
+                if dimension <= 0:
+                    dimension = self._table_shard_infos.get(
+                        table_name, _TableShardInfo()
+                    ).global_cols
+                feature_names = tuple(getattr(table_config, "feature_names", ()))
+                identity_key = (role, table_name)
+                previous_owner = owner_by_identity.get(identity_key)
+                if previous_owner is not None and previous_owner != module_fqn:
```
tiankongdeguiji: INPUT_TILE=3 legitimately splits sparse tables into ebc/ebc_user and ec_dict/ec_dict_user. If user and non-user features share a table name, these framework-created modules can expose the same (role, table_name) under different FQNs, causing this check to reject an otherwise valid input-tile layout.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+                        "inconsistent sparse embedding metadata for "
+                        f"role={role} table={table_name}: {previous} vs {current}"
+                    )
+                metadata_by_identity[identity_key] = current
+                roles_by_table.setdefault(table_name, set()).add(role)
+
+        ambiguous_tables = sorted(
+            table_name for table_name, roles in roles_by_table.items() if len(roles) > 1
+        )
+        if ambiguous_tables:
```
tiankongdeguiji: This needs to be supported rather than rejected. It is common for a sequence group’s query-side feature to also appear in a deep group, so TZRec legitimately exposes the same shared embedding table through EC for the sequence path and EBC for the pooled deep path. ambiguous_tables therefore rejects a standard model configuration, not an exceptional collision.

tzrec/utils/sparse_embedding_contract.py
```diff
@@ -0,0 +1,105 @@
...
+    name_by_identity: Dict[Tuple[str, str], str] = {}
+    for table_name, roles in roles_by_name.items():
+        if len(roles) == 1:
+            role = next(iter(roles))
+            name_by_identity[(role, table_name)] = table_name
+            continue
+
+        for role in sorted(roles):
+            base_candidate = f"{table_name}__{role}"
+            candidate = base_candidate
```
tiankongdeguiji, line 65: Could we use the cleaned table/weight FQN directly as embedding_name instead of introducing another namespace based on (role, table_name)? A table FQN already identifies the exact EC/EBC module and physical table, so it naturally avoids multiple-owner collisions, ambiguous_tables, and candidate/suffix allocation

tiankongdeguiji (changes requested):

tzrec/utils/delta_embedding_dump.py
```diff
@@ -356,20 +440,269 @@ def __init__(
...
+            logger.warning(
+                "Delta embedding dump cannot distinguish tables reused by both "
+                "EmbeddingCollection and EmbeddingBagCollection (%s); the TorchRec "
+                "tracker collapses them to one physical table, so only that "
+                "table's deltas are published.",
+                ambiguous_tables,
+            )
+
+        name_by_identity = build_sparse_embedding_name_map(metadata_by_identity)
+        identity_by_fqn: Dict[str, SparseEmbeddingIdentity] = {}
```
tiankongdeguiji: `identity` refers to both a `(role, table_name)` tuple  line 575 and a `SparseEmbeddingIdentity` line 638, which is confusing.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -381,15 +714,186 @@ def pause_tracking(self) -> Iterator[None]:
...
     def maybe_dump(self, global_step: int) -> None:
-        """Dump on the configured global-step interval and advance tracker state.
+        """Dump on the configured step or time interval and advance tracker state.
 
         Args:
             global_step: Current training step.
         """
-        if global_step > 0 and global_step % self._interval == 0:
-            self.dump(global_step)
+        if self._requires_dump_state_rendezvous():
```
tiankongdeguiji: Please dump directly when the interval is reached. There is no need to rendezvous or propagate failures from the previous step: in our runtime, a failure on any rank terminates the entire training job, so the remaining ranks will not continue or hang.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+
+def _fsync_parent_directory(path: str) -> None:
+    directory_fd = os.open(os.path.dirname(path) or ".", os.O_RDONLY)
+    try:
+        os.fsync(directory_fd)
+    finally:
+        os.close(directory_fd)
+
+
+def _durable_makedirs(path: str) -> None:
```
tiankongdeguiji: over design

tzrec/utils/delta_embedding_dump.py
```diff
@@ -758,18 +1363,30 @@ def _append_table_chunk(
...
             )
-
+        if embeddings_cpu.size(1) != expected_dimension:
+            raise ValueError(
+                f"delta embedding dimension mismatch for {embedding_name!r}: "
+                f"expected={expected_dimension}, actual={embeddings_cpu.size(1)}"
+            )
+        if not bool(torch.isfinite(embeddings_cpu).all().item()):
+            raise ValueError(f"delta embedding {embedding_name!r} contains NaN or Inf")
+        if bool((key_ids_cpu == SPARSE_EMBEDDING_INVALID_KEY).any().item()):
```
tiankongdeguiji: no needed to check

tzrec/utils/delta_embedding_dump.py
```diff
@@ -758,18 +1363,30 @@ def _append_table_chunk(
                 "delta embedding dump key ids and embeddings row count mismatch: "
                 f"key_ids={num_rows}, embeddings={embeddings_cpu.size(0)}."
             )
-
+        if embeddings_cpu.size(1) != expected_dimension:
+            raise ValueError(
+                f"delta embedding dimension mismatch for {embedding_name!r}: "
+                f"expected={expected_dimension}, actual={embeddings_cpu.size(1)}"
+            )
+        if not bool(torch.isfinite(embeddings_cpu).all().item()):
```
tiankongdeguiji: no need to check

tzrec/utils/delta_embedding_dump.py
```diff
@@ -792,19 +1409,36 @@ def _embedding_array(self, embeddings: torch.Tensor) -> pa.ListArray:
...
+                writer_schema = writer_schema.with_metadata(metadata)
+            with pq.ParquetWriter(tmp_path, writer_schema) as writer:
                 chunks = table_chunks or [_DELTA_DUMP_SCHEMA.empty_table()]
                 for table_chunk in chunks:
                     writer.write_table(table_chunk)
+            with open(tmp_path, "rb") as source:
+                os.fsync(source.fileno())
             os.replace(tmp_path, output_path)
+            directory_fd = os.open(os.path.dirname(output_path) or ".", os.O_RDONLY)
+            try:
```
tiankongdeguiji: Review PR #603 under the following product and recovery contract:

1. Delta embedding dumps produced after the latest checkpoint are speculative.
2. If the training process fails or exits abnormally, training restarts from the latest checkpoint.
3. Any pending delta generated after that checkpoint is considered stale and must be discarded.
4. We do not need to replay pending uploads across process restarts.
5. We do not require local upload rollback or remote undo.
6. FeatureStore upload is best-effort for the current live training process.
7. Retry within the same live process is acceptable, but cross-restart recovery is explicitly out of scope.

Under this contract, review the PR specifically for over-engineered durability and crash-recovery mechanisms.

Identify every place that implements a durable outbox or cross-restart replay, including but not limited to:

- file and directory fsync
- durable directory creation
- persistent upload journals
- contract.json / committed.json / success markers
- persistent error markers
- uploader-owned shard snapshots
- shard content hashing for restart recovery
- manifest attempt history
- persisted timestamp reservations
- startup discovery and reconciliation of historical steps
- replay of uncommitted steps after restart
- superseding the same global step with a new dump generation
- tracker durability consumers and durable acknowledgement cursors
- writer locks whose only purpose is cross-process recovery
- cleanup of stale snapshot staging directories
- offline tools that depend on persistent commit markers

Evaluate whether the implementation can be simplified to this lifecycle:

1. Each rank writes one temporary Parquet shard.
2. The completed shard is atomically renamed into place; no fsync durability guarantee is required.
3. Rank zero receives the completed step through an in-memory queue.
4. Rank zero waits for the complete shard set, validates it, streams/deduplicates it and uploads it.
5. A successful upload removes or ignores the local files.
6. An abnormal process exit abandons all pending files and in-memory state.
7. A restarted training run does not scan, reconcile or replay artifacts from the previous run.


Pay particular attention to:

- tzrec/utils/feature_store_delta_uploader.py
- tzrec/utils/delta_embedding_dump.py
- tzrec/tools/feature_store/check_feature_store_delta.py
- tzrec/protos/train.proto

tiankongdeguiji (changes requested):

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+    while True:
+        chunk = source.read(8 * 1024 * 1024)
+        if not chunk:
+            break
+        digest.update(chunk)
+    source.seek(0)
+    return digest.hexdigest()
+
+
+class _RecordStore:
```
tiankongdeguiji: Could we avoid funneling all rank shards through the rank-zero SQLite dedup store? For table-wise and row-wise sharding, `(embedding_name, key_id)` already has a unique owner, so the extra snapshot read, SQLite write, dedup, and sort add substantial I/O without changing the result. For data-parallel tables, each rank could upload its shard independently; duplicate `MERGE` writes are acceptable when replicas are consistent, and DP can be documented as discouraged for large tables. Please consider direct per-rank streaming uploads, coordinating only FeatureView creation, timestamp allocation, and global success/failure. If centralized replay is required specifically for strict Next-Ts ordering, that invariant and trade-off should be documented; otherwise this machinery appears unnecessarily complex.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+        with self._condition:
+            if self._started:
+                return
+            if self._closed:
+                raise RuntimeError("FeatureStoreDeltaUploader is already closed")
+            self._raise_if_failed_locked()
+            self._acquire_writer_lock()
+            try:
+                self._initialize_journal()
+                self._cleanup_stale_snapshot_staging()
```
tiankongdeguiji: This path writes the already-materialized delta to Parquet, copies it into a full snapshot, reads it back, spills it through SQLite, and then materializes SDK batches again. Please consider a bounded direct-memory fast path for the active process.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+                        )
+                    else:
+                        try:
+                            snapshot_paths = self._snapshot_canonical_shards(
+                                current_step, canonical_paths
+                            )
+                        except _ShardSetNotReady:
+                            snapshot_paths = None
+
+                if snapshot_paths is None:
```
tiankongdeguiji: do not need snapshot

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,2305 @@
...
+            "contract_hash": self._contract_hash,
+            "record_count": record_count,
+            "dump_generation": dump_generation,
+            "shards": expected_shards,
+            "attempts": [],
+        }
+        _atomic_write_json(path, manifest)
+        return manifest
+
+    def _start_attempt(
```
tiankongdeguiji: Persistence and restoration are meaningless—the training resumes from the previous checkpoint.

tiankongdeguiji (changes requested):

tzrec/utils/sparse_embedding_contract.py
eric-gecheng: Thanks — declined the FQN switch and instead deduped the algorithm in f62d861.

`embedding_name` is a cross-system serving key, not a delta-local value: the contract module docstring states it is shared by export, delta dump, and serving. Concretely, the export path (`export_util.py`) derives the name from `BaseEmbeddingConfig.name + role` via the identical role-suffix algorithm and writes it into `feat_meta["embedding_name"]`, which serving/inference reads to query the FeatureStore; the delta side sends that same string as the remote MERGE primary-key value (`FEATURE_STORE_PK_FIELD = "embedding_name"`). The export path has no module FQN to align a delta-side FQN with, so switching only the delta side would diverge from serving and orphan every delta row. A bare `module_fqn` would also collapse a multi-table EBC module onto one primary key and trip the dimension consistency check.

Also, the premise that the current scheme needs FQN to avoid collisions isn't quite right: `build_sparse_embedding_name_map` already disambiguates EC/EBC (each gets `f"{table_name}__{role}"` when both collections share a name), and the old `ambiguous_tables` guard was removed in 877ece7.

The real duplication was that `export_util.py` kept its own private copy of the algorithm. f62d861 drops the private `_SPARSE_EC_ROLE`/`_SPARSE_EBC_ROLE` constants and the `_build_sparse_export_name_map`/`_resolve_sparse_export_name` copies and imports from `sparse_embedding_contract` instead, so the two sides can't drift. A differential fuzz over 610 cases confirms byte-identical output for every realistic config; the only behavior change is raising on an empty table name (impossible for real configs via the `embedding_name or f"{name}_emb"` fallback).

tzrec/utils/delta_embedding_dump.py
eric-gecheng: Done in 272fdb8 — renamed the tuple-keyed `table_identities` / `name_by_identity` to `role_table_pairs` / `name_by_role_table` so the key shape is explicit, and added a one-line note that `identity_by_owner`'s key is `(module_fqn, table_name)`, distinct from the contract's `(role, table_name)` key. Kept the `SparseEmbeddingIdentity` dataclass and its instance locals; they are the legitimate singular. Pure rename, all callers are positional, behavior and the exported embedding_name are unchanged.

tzrec/utils/delta_embedding_dump.py
eric-gecheng: no impact

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+            allow_custom_endpoint=allow_custom_endpoint,
+        )
+
+
+def validate_feature_store_config(config: FeatureStoreConfig) -> None:
+    """Validate upload configuration without resolving credentials."""
+    FeatureStoreUploadSettings.from_proto(config)
+
+
+def _validate_feature_store_endpoint(
```
tiankongdeguiji: This appears over-designed for a user-supplied endpoint. Both `endpoint` and `allow_custom_endpoint` are controlled by the same configuration author, so neither the hostname allowlist nor its opt-out establishes a meaningful security boundary.

Endpoint selection is an explicit user action, and the FeatureStore SDK already owns endpoint handling and connection errors. Please remove `allow_custom_endpoint` and `_validate_feature_store_endpoint`, and pass the configured endpoint directly to the SDK.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -85,19 +114,26 @@ def validate_delta_embedding_dump_config(
...
+                "dump_interval_steps and dump_interval_minutes."
+            )
+        if config.dump_interval_minutes <= 0:
+            raise ValueError(
+                "delta_embedding_dump_config.dump_interval_minutes must be > 0."
+            )
+    elif config.dump_interval_steps <= 0:
         raise ValueError("delta_embedding_dump_config.dump_interval_steps must be > 0.")
+    if config.HasField("feature_store_config"):
+        validate_feature_store_config(config.feature_store_config)
```
tiankongdeguiji: will be validated in FeatureStoreDeltaUploader again, do not need to validate_feature_store_config here, also i think validate_feature_store_config the function is not needed.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+    ) -> None:
+        """Initialize the uploader with validated settings and in-memory state."""
+        self._settings = FeatureStoreUploadSettings.from_proto(config)
+        self._rank = int(rank)
+        self._manage_remote_view = bool(manage_remote_view)
+        self._embedding_dimensions = {
+            str(name): int(dimension)
+            for name, dimension in embedding_dimensions.items()
+        }
+        invalid_dimensions = {
```
tiankongdeguiji: Please remove the `invalid_dimensions` check. Excessively defensive programming

tiankongdeguiji (changes requested):

tzrec/utils/delta_embedding_dump.py
```diff
@@ -447,30 +730,51 @@ def dump(self, global_step: int) -> Optional[str]:
...
-            dynamic_modules=dynamic_modules,
-        )
+        global_step = int(global_step)
+        if global_step <= 0:
+            raise ValueError("delta embedding dump global_step must be > 0")
+        uploader = self._uploader
+        write_local = not self._feature_store_enabled or self._retain_local_dump
+        tracker_cursor = self._tracker_cursor_before_read()
+        try:
+            table_weights = self._collect_table_weights()
```
tiankongdeguiji: Could we discover and cache table_weights and dynamic_modules lazily on the first dump instead of traversing all tracked modules on every dump?

tzrec/utils/delta_embedding_dump.py
```diff
@@ -447,30 +730,51 @@ def dump(self, global_step: int) -> Optional[str]:
...
+            output_path: Optional[str] = None
+            if write_local and (num_rows > 0 or self._world_size > 1):
+                # Multi-rank shard sets stay complete even for an empty rank so
+                # per-step file consumers never observe a partial set.
+                output_path = self._output_path(global_step)
+                self._write_table_chunks(table_chunks, output_path)
+            if uploader is not None and num_rows > 0:
+                uploader.submit(global_step, pa.concat_tables(table_chunks))
+        except BaseException:
+            self._rollback_tracker_read(tracker_cursor)
```
tiankongdeguiji: This rollback is unnecessary and ineffective: get_unique() deletes consumed rows, so restoring only the cursor cannot recover them. Please remove _tracker_cursor_before_read() and _rollback_tracker_read().

tzrec/utils/delta_embedding_dump.py
```diff
@@ -404,39 +684,42 @@ def final_dump(self, global_step: int) -> Optional[str]:
...
+            # already-written boundary shards (with an empty file under
+            # multi-GPU), so skip.
+            return None
+        if self._interval_secs is not None and global_step == self._last_dump_step:
+            # A timed dump can land on any step. Avoid replacing that step's full
+            # delta with an empty final shard when training ends immediately after.
             return None
         return self.dump(global_step)
 
     def _sync_final_step(self, global_step: int) -> int:
```
tiankongdeguiji, line 750: Final steps should already be aligned across ranks; otherwise distributed training would hang or fail before reaching `final_dump()`. The extra `MAX` all-reduce is unnecessary. Please remove `_sync_final_step()` and use the local `global_step` directly.

tiankongdeguiji (changes requested):

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+    because a key's owner rank is fixed for the lifetime of the process.
+    """
+
+    def __init__(
+        self,
+        config: FeatureStoreConfig,
+        embedding_dimensions: Mapping[str, int],
+        rank: int = 0,
+        manage_remote_view: bool = True,
+        client_factory: Optional[Callable[..., Any]] = None,
```
tiankongdeguiji: Could we remove both client_factory and credentials_client from __init__ and encapsulate credential resolution and FeatureStoreClient construction in a private _create_client() method? Tests can patch that single method instead.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -354,22 +413,213 @@ def __init__(
...
+            if self._world_size > 1:
+                if not (
+                    torch.distributed.is_available()
+                    and torch.distributed.is_initialized()
+                ):
+                    raise RuntimeError(
+                        "distributed FeatureStore delta dump requires an "
+                        "initialized process group"
+                    )
+                torch.distributed.barrier()
```
tiankongdeguiji: Could we move this barrier into `FeatureStoreDeltaUploader`? The rank-zero create/validate → barrier → other ranks open sequence is part of the uploader's startup protocol. Keeping it in `DeltaEmbeddingDumper` leaks the `manage_remote_view` implementation details to the caller.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+                )
+            self._view = view
+            return view
+        provisioned = False
+        view = project.get_dynamic_embedding_feature_view(
+            self._settings.feature_view_name
+        )
+        if view is not None:
+            self._view = view
+        metadata = self._wait_for_feature_view_metadata(project)
```
tiankongdeguiji: This additional control-plane `FeatureView` lookup and validation seems over-designed. `get_dynamic_embedding_feature_view()` already performs `ListFeatureViews` and `GetFeatureView` internally before constructing the data-plane writer, so `_wait_for_feature_view_metadata()` fetches the same resource again solely to enforce metadata and exact provisioning values.
TTL, shard count, and replication count are creation-time settings and do not determine upload compatibility. The extra lookup also introduces redundant API calls and another eventual-consistency failure path. Please rely on the `DynamicEmbeddingFeatureView` PK/SK/embedding mapping and remove this control-plane metadata wait. If field-type metadata is genuinely required, it should be preserved and exposed by the SDK from the original `GetFeatureView` response instead of being fetched twice.

tiankongdeguiji (changes requested):

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+                )
+            if self._closing or self._closed:
+                raise RuntimeError("cannot submit to a closing FeatureStore uploader")
+            while (
+                global_step not in self._pending
+                and len(self._pending) >= self._settings.max_pending_steps
+            ):
+                self._condition.wait(self._settings.poll_interval_secs)
+                self._raise_if_failed_locked()
+            self._pending.setdefault(global_step, table)
```
tiankongdeguiji: Could we use collections.deque for _pending instead of a dictionary? This is an in-process FIFO queue, and submissions already arrive in training order, so keying entries by global_step and calling min(self._pending) are unnecessary.

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+        range_start = ts_range[0]
+
+        completed_batches = 0
+        window_batches = 0
+        window_records = 0
+        started_at = time.monotonic()
+        next_progress_batch = _FEATURE_STORE_PROGRESS_LOG_INTERVAL_BATCHES
+        logged_first_window = False
+
+        logger.info(
```
tiankongdeguiji: `_stream_upload` emits start, first-window progress, and completion messages at INFO for every rank and dump. For small deltas these messages are largely redundant, and the volume scales with world size and dump frequency. Could we log startup and the first window at DEBUG, while keeping periodic progress and the final completion at INFO?

tzrec/utils/delta_embedding_dump.py
```diff
@@ -447,30 +730,51 @@ def dump(self, global_step: int) -> Optional[str]:
...
+                logger.info(
+                    "Dumped empty delta embedding shard to %s at step %s.",
+                    output_path,
+                    global_step,
+                )
+        elif output_path is None:
             logger.info(
-                "Dumped empty delta embedding shard to %s at step %s.",
-                output_path,
+                "Submitted %s delta embedding rows for FeatureStore upload at step %s.",
```
tiankongdeguiji, line 815: use debug logger level or log with intervals

tzrec/utils/feature_store_delta_uploader.py
```diff
@@ -0,0 +1,942 @@
...
+        bad_rows = np.flatnonzero(lengths != expected_dims)
+        if bad_rows.size > 0:
+            row = int(bad_rows[0])
+            raise ValueError(
+                f"delta embedding dimension mismatch for {embedding_names[row]!r}: "
+                f"expected={int(expected_dims[row])}, "
+                f"actual={int(lengths[row])}"
+            )
+
+        return [
```
tiankongdeguiji, line 648: This conversion defeats the columnar pipeline and will become the upload bottleneck for large deltas. Each Arrow batch is expanded into one Python dict and one copied NumPy array per row; feature_store_py.write_features() then rebuilds every dict, calls ndarray.tolist(), and finally serializes the result with json.dumps(). This creates O(rows × embedding_dim) Python objects and performs multiple full-data copies before every HTTP request.Please address this at the SDK boundary instead of optimizing the List[Dict] construction locally. feature_store_py should expose a public columnar/binary bulk-write API accepting a pyarrow.RecordBatch or equivalent PK/SK/embedding buffers, validate them vectorially, and encode them directly into the wire payload without materializing per-element Python objects. Prefer FlatBuffers/Arrow IPC.

tiankongdeguiji (changes requested):

tzrec/main.py
```diff
@@ -835,25 +869,60 @@ def train_and_evaluate(
...
-        eval_config=pipeline_config.eval_config,
-        ckpt_manager=ckpt_manager,
-        skip_steps=skip_steps,
-        ckpt_path=ckpt_path,
-        check_all_workers_data_status=check_all_workers_data_status,
-        ignore_restore_optimizer=ignore_restore_optimizer,
-        dataloader_state=dataloader_state,
-        delta_embedding_dumper=delta_embedding_dumper,
-    )
+    try:
```
tiankongdeguiji: No need to catch training failures here. `torchrun` terminates the job, and pending deltas are intentionally discarded. Only close the dumper after successful training to drain uploads.

tiankongdeguiji (changes requested):

tzrec/utils/feature_store_delta_uploader.py
eric-gecheng: using arrow instead

tiankongdeguiji (approved):

PR #604, opened 2026-07-16
[feat]online export of dense models

eric-gecheng:


tzrec/utils/env_util.py
```diff
@@ -32,6 +32,11 @@ def use_rtp() -> bool:
     return flag
 
 
+def use_distributed_embedding() -> bool:
```
tiankongdeguiji: use_distributed_embedding already in acc/utils.py

tiankongdeguiji (changes requested):

tzrec/utils/online_dense_export_util.py
```diff
@@ -0,0 +1,727 @@
...
+        snapshot: Dict[str, torch.Tensor] = {}
+        if self._state_pairs:
+            source_state = model.state_dict()
+            for gm_key, source_key in self._state_pairs:
+                value = source_state[source_key]
+                if isinstance(value, DTensor):
+                    # collective on the DTensor's mesh; all ranks participate
+                    gathered = value.full_tensor()
+                elif isinstance(value, ShardedTensor):
+                    gathered = self._gather_sharded_tensor(value)
```
tiankongdeguiji: i think we should not gather_sharded_tensor in dense export

tzrec/utils/online_dense_export_util.py
```diff
@@ -0,0 +1,727 @@
...
+                elif isinstance(value, torch.Tensor):
+                    gathered = value
+                else:
+                    raise RuntimeError(
+                        f"ONLINE_DENSE_EXPORT cannot gather dense state "
+                        f"[{gm_key}]: unsupported state type "
+                        f"{type(value).__name__}"
+                    )
+                if is_rank_zero:
+                    snapshot[gm_key] = gathered.detach().cpu()
```
tiankongdeguiji: [P2] gathered.detach().cpu() synchronizes the training thread for every CUDA tensor, so the supposedly background dense export can introduce significant step latency for large models. Consider copying all CUDA tensors into pinned CPU buffers with non_blocking=True, then synchronizing the copy stream once before enqueueing the snapshot. A distributed barrier is unnecessary here; CUDA stream/event synchronization is what guarantees that the worker sees completed copies.

tzrec/tools/online_dense_export.py
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    # Keep the service-facing pointer beside the immutable dense export versions.
+    _publish_current(os.path.join(export_root, CURRENT_JSON), current_payload)
+    _prune_old_dense_versions(export_root, versions_root)
+    logger.info("published online dense export version %s to %s", version, version_dir)
+    return current_payload
+
+
+def main() -> None:
```
tiankongdeguiji, line 147: the tool is still needed or not？

tiankongdeguiji (changes requested):

tiankongdeguiji:
bump up version to 1.3.6

tiankongdeguiji (approved):

PR #622, opened 2026-08-03
[feat] convert zch to dynemb in distributed export

eric-gecheng:


tzrec/utils/export_util.py
```diff
@@ -2307,6 +2313,130 @@ def _prepare_sparse_export_values(
...
+    # the row MCHManagedCollisionModule.remap() sends unmatched ids to.
+    default_row = mch._output_global_offset + mch._zch_size - 1 - shard_offset
+    if default_row < 0 or default_row >= num_rows:
+        raise ValueError(
+            f"zch table {emb_name} default row {default_row} is out of its "
+            f"{num_rows} embedding rows at shard offset {shard_offset}."
+        )
+    return (
+        keys,
+        local_tensor[rows.numpy()],
```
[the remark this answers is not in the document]
eric-gecheng: Valid — fixed in be25da9. The caller now passes the torch weight (still on device) down, and `_zch_table_to_dynamic` gathers the selected rows on the weight's device with `index_select` before the single `.cpu().numpy()", so only the exported rows (plus the one default row) reach host; the full-table host copy no longer happens for zch tables. Non-zch tables keep their single host copy as before.

tzrec/utils/export_util.py
```diff
@@ -2307,6 +2313,135 @@ def _prepare_sparse_export_values(
...
+    raw_ids = mch._buffers["_mch_sorted_raw_ids"].to(weight.device)
+    remapped_ids = mch._buffers["_mch_remapped_ids_mapping"].to(weight.device)
+    # LFU keeps counts only, LRU keeps last access iter only, DistanceLFU keeps
+    # both and its recency part maps onto the dynamic table score.
+    score_buffer = None
+    for buffer_name in ("_mch_last_access_iter", "_mch_counts"):
+        score_buffer = mch._buffers.get(buffer_name)
+        if score_buffer is not None:
+            break
+    if score_buffer is None:
```
tiankongdeguiji: Export and serving do not strictly require score semantics. When score_buffer is None, could we emit an int64 zero score for each valid key instead of raising an exception? Existing eviction scores should still be preserved when the metadata is available. Please also add a test for this fallback and document the behavior.

docs/source/feature/zch.md
```diff
@@ -27,6 +27,14 @@ feature_configs {
 
 - **threshold_filtering_func**: 准入策略lambda函数，默认为全部准入，详见下文准入策略
 
+## 分布式 Embedding 导出
```
tiankongdeguiji, line 30: Do not need the doc. we describe only the user-observable behavior here. Artifact names and fields such as sparse_dynamic_embedding-*.npz, sparse_embedding.json, is_dynamic, and default_value are internal export/serving contracts and may change independently.

tzrec/modules/embedding.py
```diff
@@ -1260,15 +1300,23 @@ def _forward_impl(
...
                 sparse_jt_dict_list.append(new_d_jt)
 
         if self.has_mc_sparse_user:
-            for ec in self.mc_ec_dict_user.values():
-                sparse_jt_dict_list.append(ec(sparse_feature_user)[0])
+            for feature_keys, mc_ec in zip(
+                self.mc_ec_dict_features_user.values(), self.mc_ec_dict_user.values()
+            ):
+                d_jt = mc_ec(sparse_feature_user)[0]
+                new_d_jt = {}
```
tiankongdeguiji: Could we extract this repeated marking and key-remapping logic into a helper? It is currently duplicated across four paths.

tiankongdeguiji (changes requested):
bump up to 1.3.11

tiankongdeguiji (approved):

PR #625, opened 2026-08-05
[feat] prompt-native generative recommendation

WhiteSwan1:
## Summary

This PR adds prompt-native generative recommendation to TorchEasyRec. `PromptConfig` owns prompt composition, response declaration, tokenizer extension, SID space, and INLINE/PROJECTED slots, while `GenrecCausalLMModel` owns causal-LM training and band-constrained SID decoding.

The main changes are:

- Compile `prompt_config` into a prompt plan, projection plan, resolved SID space, and checkpoint digests while resolving the extended SID vocabulary.
- Assemble each parsed batch into flat token ids, row offsets, projected-slot positions, response lengths, and a host-computed maximum sequence length. Assembly loops over the fixed segment plan rather than over samples.
- Read the response exclusively from `data_config.label_fields`. It remains declared in every mode, but predict-mode assembly omits it from the rendered prompt.
- Support INLINE SID sequences and PROJECTED feature slots. PROJECTED slots derive their feature groups and map their embeddings into the LM hidden size before replacing sentinel positions.
- Add `BaseGenrecModel` and `GenrecCausalLMModel`. The model resizes the HF vocabulary, loads pretrained weights only for a cold start, returns only the bounded response-window logits for training, delegates CE computation to the backbone, and decodes fixed-width SIDs with per-level token bands and dynamic beam widths.
- Record the backbone state-dict prefix plus prompt digests beside DCP checkpoints. Restore rejects a missing or mismatched vocabulary digest and warns when only the prompt plan changes. If rank 0 cannot write the HF assets, the error is broadcast to all ranks and the save fails before writing dataloader state, metadata, `_SUCCESS`, or scheduling pruning.
- Dispatch export by the resolved model class. The current GenRec converter produces an HF backbone plus the regenerated extended tokenizer for INLINE-only prompts and rejects Dense EMA or PROJECTED prompts instead of silently dropping state.
- Add a linear-decay learning-rate scheduler with optional warmup.
- Declare `transformers` without a version pin or repository-specific wheel URL.

## Motivation

Collision output already provides `offset_codebook`. The prompt-native path consumes those offset SID codes directly, validates each level against the configured codebook band, and performs only the uniform shift by the base tokenizer vocabulary size.

This also separates responsibilities: feature extraction remains in `feature_configs`, the answer is a label, prompt ordering and literal text live in `prompt_config`, and the model receives a compiled plan instead of reinterpreting configuration at runtime.

## Design choices

- The compiler stores no physical projection dimensions. Input dimensions come from the derived feature groups and the output dimension comes from the instantiated LM, so the same compiled plan is not tied to one hidden size.
- `init_backbone()` stays on `BaseGenrecModel` because the base class also owns vocabulary resizing, input embedding lookup, projection sizing, loss, cold-start loading, and checkpoint/export access to the backbone.
- `vocab_hash` mismatch is fatal because it changes the token rows addressed by the SID bands. `plan_hash` mismatch remains a warning so an explicit fine-tune or partial restore may intentionally change prompt layout or projection topology.
- Full-vocabulary FP32 normalization is retained during beam search so candidates from different parent beams remain comparable. A chunked or fused memory optimization is separate from this change.

## Current scope

- The dataloader assembler emits a packed flat representation, but the current HF compatibility adapter left-pads it before the LM forward. Native packed-varlen LM execution is not part of this PR.
- The assembler remains a NumPy dataloader component. A shared Tensor/TorchScript assembler requires the exported serving front end to consume the same module and is deferred with that export work.
- GenRec export currently converts only the HF backbone and regenerates the tokenizer. It does not restore and export sparse embeddings, projection parameters, or a scripted prompt front end; PROJECTED prompts therefore fail explicitly. The serving-side `cache_ids` contract is also deferred with that front end.
- `sid_space` is required because this version only implements SID-generating prompt models.
- Full end-to-end multi-rank GenRec training and serving are not claimed by this change.

## Test Plan

- Repository CI covers Build Test CI, Unit Test H20 CI, PyTyping CI, Code Style CI, Unit Test CI, and Unit Test CPU CI.
- Prompt compilation, batch assembly, response-label handling, projection construction, checkpoint digest checks, GenRec forward/loss/decode, dynamic beam search, HF conversion, custom-model export dispatch, and linear-decay scheduling are covered by unit or integration tests in this PR.
- `PYTHONPATH=. python -m unittest tzrec.utils.checkpoint_util_test tzrec.prompt.persist_test` (84 tests passed for the checkpoint follow-up)
- `bash scripts/gen_proto.sh`
- `pre-commit run -a`
- `pyrefly check`


WhiteSwan1:
Thanks — this was a useful pass. Triage of all 12 comments below: **3 fixed**, **6 deferred**, **3 not taken**. Where the behaviour was checkable I verified it against the code rather than judging by reading; the evidence is in each item below.

## Fixed

- **`plan_hash` omitted projection routing** (`compile.py`) — `sorted(projection_plan.projections)` iterates a dict, so the digest covered only the module ids. `plan_hash` now hashes both `projections.items()` and `slot_to_module.items()`. Without the latter, two slots could swap `projection_name` with matching bodies and still hash the same, so the keys load and each slot receives the other's learned weights. Two regression tests, both of which fail if either mapping is dropped.
- **Missing prompt assets failed open** (`persist.py`) — agreed, and it is worse in combination with `CheckpointManager.save()` swallowing rank-0 asset-write failures. An absent `prompt_hashes.json` is now fatal. No legacy bypass, since this stack has never been released.
- **The inference tracer could not see the projected lookup** (`prompt_generative_qwen.py`) — correct, and the asymmetry was unintentional: the training path already built embeddings outside the leaf. `predict` now does so on both paths and passes them in, leaving only the padding and the LM hidden.

## Deferred, with reasons

- **HF export drops projected-slot state** — real, and tracked. It affects only a projected-slot model; the INLINE-only SID configuration this PR targets exports correctly, which is what the smoke runs and the checked-in mock config exercise. Taking the "fail export when projected slots exist" half separately is attractive, since `dcp_to_hf` currently writes a *valid* `model.safetensors` containing only the backbone, so the incompleteness is silent.
- **Response plan vs. banded decode** — accepted as a real train/serve mismatch, but nothing is broken for the shape in use: a response holding one INLINE SID slot trains and decodes consistently. Restricting the plan now would forbid shapes nobody has requested; better resolved alongside the extension that wants them.
- **SID ids placed without knowing the backbone's table** — the sharpest comment here, and it goes deeper than stated. Verified: SID ids land *inside* the backbone's existing table while the resize simultaneously adds rows above it that no token addresses, and a small `sum(codebook)` can make `target_vocab_size` round *below* `config.vocab_size` and shrink the table. A `base_vocab_size == config.vocab_size` check is not the fix — HF pads above the tokenizer (Qwen2.5 ships `vocab_size 151936` against ~151.6k tokens), so that would reject every legitimate config. The correct invariant is `max(tokenizer_vocab_size, lm.config.vocab_size)`, but `compile_prompt` has no access to the backbone config and the bands are already hashed into `vocab_hash` by the time the model could compare. That makes it a migration, not a patch, and it wants validating against a real backbone rather than a 64-vocab fixture.
- **`max_seqlen` read back off the device** — confirmed: `Batch.to()` migrates every `additional_infos` tensor, so this is a device-to-host sync per step, and it contradicts a design property the surrounding comments asserted. Performance only, though — the width is the same number either way. A fix was written and reverted: it carried the width as a plain `int` field on `Batch` alongside `tile_size`, which worked, but adds a prompt-specific field to a dataclass every model touches. Deferred pending a GPU measurement to size the win. The stale docstring claiming the model never reads it off the device has been corrected regardless.
- **Full-vocabulary fp32 logits per beam** and **per-beam KV cache duplication** — both real and both want a design rather than a patch (chunked or fused `lm_head` + log-sum-exp; prefix-sharing or a paged cache). Worth noting the current `_band_logp` is already a deliberate trade — normalize then slice, rather than a full-vocab `log_softmax` — so these ask for the next step rather than correct an oversight. The KV duplication is likely the largest single inference-memory item.

## Not taken

These three each guard a state the pipeline does not produce:

- **Response width unchecked per row** — every route was checked. A codebook mismatch is already caught by the offset validation (`_inline_tokens` rejects values that do not carry their level offset — the self-validating property the offset encoding exists for). An answer holding two items requires an upstream labeling bug. An empty answer cannot come from the SID pipeline: `resolve_sid_collisions` never drops an item — one it cannot relocate keeps its original SID and overflows the bucket.
- **Non-finite / fractional SID values** — same guarantee. SID codes arrive from the collision tool's `offset_codebook` column as integers.
- **Static-only prediction prompt assembles zero rows** — reproduced, but a prompt whose body carries no slot has no per-request content at all; every request would produce the same static prefix. Reaching that state means the prompt needs rethinking, not that the assembler needs a row count.

The first two rest on one upstream guarantee — the `offset_codebook` column supplies exactly one complete SID per row, as integers — which is real but stated nowhere in the prompt stack. Noted as an assumption to make explicit.

tzrec/modules/dynamic_beam.py
```diff
@@ -0,0 +1,134 @@
...
+            because every use of them is a host-side slice bound.
+
+    Returns:
+        The SID token tail ``(B * W, num_levels)``, score-ordered best-first.
+        The fixed-length, EOS-free answer needs no finished-sequence bookkeeping.
+    """
+    device = prompt_embeds.device
+    batch_size = prompt_embeds.shape[0]
+    num_levels = len(bands)
+    if len(beam_widths) != num_levels:
```
tiankongdeguiji: These values are static for the lifetime of the model. The beam schedule is already validated in _read_beam_config(), and both bands and the capped widths depend only on the compiled SID space and model configuration. Please compute and store them during initialization instead of rebuilding and revalidating them on every inference call.

tzrec/main.py
```diff
@@ -1078,6 +1111,53 @@ def export(
...
+        if (
+            pipeline_config.HasField("export_config")
+            and pipeline_config.export_config.exporter_type == "best"
+        ):
+            checkpoint_path, _ = ckpt_manager.best_checkpoint()
+        else:
+            checkpoint_path, _ = ckpt_manager.latest_checkpoint()
+
+    # HF export converts the checkpoint dir directly -- no model build, no DCP restore.
+    if pipeline_config.export_config.export_format == export_pb2.ExportFormat.HF:
```
tiankongdeguiji: The HF export path should still build and restore the full model. Direct DCP-to-HF conversion only exports the backbone and drops sparse embeddings and projection parameters, which must be included in the scripted model. Therefore, storing HF assets in every checkpoint is redundant; generate them from the restored model during export instead.

tzrec/protos/export.proto
```diff
@@ -1,6 +1,11 @@
 syntax = "proto2";
 package tzrec.protos;
 
+enum ExportFormat {
```
tiankongdeguiji: export_format also seems unnecessary here because the supported export path is determined by the model type. The exporter should dispatch based on model capabilities

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+from tzrec.utils.logging_util import logger
+
+_PARAM_DTYPE: Dict[int, torch.dtype] = {
+    PromptModelConfig.FP32: torch.float32,
+    PromptModelConfig.BF16: torch.bfloat16,
+    PromptModelConfig.FP16: torch.float16,
+}
+
+
+class BasePromptGenerativeModel(BaseModel):
```
tiankongdeguiji: Name this abstraction after GenRec, not prompt rendering. Prompt compilation belongs to the input pipeline; this class owns shared generative-recommendation behavior. Rename the module to genrec_model.py and the class to BaseGenrecModel.

tzrec/models/prompt_generative_qwen.py
```diff
@@ -0,0 +1,267 @@
...
+    PROMPT_INPUT_IDS,
+    PROMPT_MAX_SEQLEN,
+    PROMPT_RESPONSE_LENGTHS,
+)
+from tzrec.prompt.plan import CompiledPrompt
+from tzrec.protos.model_pb2 import ModelConfig
+from tzrec.protos.models.prompt_model_pb2 import PromptModelConfig
+
+
+class PromptGenerativeQwen(BasePromptGenerativeModel):
```
tiankongdeguiji: avoid encoding Qwen in the model identity. This implementation constructs an arbitrary AutoModelForCausalLM; Qwen is only the currently tested backbone. Rename the module to genrec_causal_lm_model.py and the class to GenrecCausalLMModel, then validate the required causal-LM interfaces explicitly.

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+    def init_input(self) -> None:
+        """Build the projected-slot embedding groups and projection modules."""
+        self.embedding_group = EmbeddingGroup(
+            self._features, list(self._prompt.projection_plan.feature_groups)
+        )
+        self._build_projections()
+
+    def _build_backbone(
```
tiankongdeguiji: Make backbone construction an initialization step. Move this method to GenrecCausalLMModel, rename it to init_backbone(), assign self.lm inside it, and return None so its lifecycle matches init_input() and init_from_pretrained().

tzrec/models/model.py
```diff
@@ -94,6 +94,25 @@ def predict(self, batch: Batch) -> Dict[str, torch.Tensor]:
         """
         raise NotImplementedError
 
+    def save_assets(self, target_dir: str) -> None:
```
tiankongdeguiji: Remove the model-level asset hook. Prompt compilation artifacts are owned by the pipeline and export flow, not by model weights. The generic BaseModel API should not gain a lifecycle method used only to persist prompt metadata.

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+    ) -> None:
+        """No-op: nothing beyond the logged CE.
+
+        Args:
+            predictions: what ``predict`` returned.
+            batch: the batch, unused.
+        """
+        return
+
+    def save_assets(self, target_dir: str) -> None:
```
tiankongdeguiji: Do not make the model persist CompiledPrompt. pipeline.config is the prompt source of truth, while serving executes the scripted front-end and does not ask the model to reconstruct or publish the prompt contract. Remove this override.

tzrec/prompt/persist.py
```diff
@@ -0,0 +1,160 @@
...
+            f.name: _plain(getattr(value, f.name)) for f in dataclasses.fields(value)
+        }
+    if isinstance(value, (list, tuple)):
+        return [_plain(v) for v in value]
+    if isinstance(value, dict):
+        return {str(k): _plain(v) for k, v in value.items()}
+    return value
+
+
+def save_prompt_assets(compiled_prompt: CompiledPrompt, target_dir: str) -> None:
```
tiankongdeguiji: Do not persist PromptPlan as a serving artifact. RFC 0001/0002 compile the walk into the scripted front-end specifically to avoid a second serving-side assembler. Keep any readable plan dump as an optional diagnostic report, not a per-checkpoint asset.
Do not publish a standalone sid_space.json. Serving obtains token identity from the extended tokenizer, legal SID paths and item resolution from the constraint index, and runtime settings from deployment configuration. This file has no runtime reader and duplicates pipeline.config.
Do not create a separate hash artifact. Embed the contract digest in the HF metadata, scripted front-end, and constraint-index metadata so the loader verifies the artifacts it actually consumes. A standalone JSON only adds another file that can drift.

tzrec/utils/checkpoint_util.py
```diff
@@ -390,9 +418,34 @@ def save(
...
+
+        # a raise here skips save_dataloader_state's all_gather and hangs other ranks.
+        try:
+            write_hf_assets(model, ckpt_dir)
+        except Exception as e:  # noqa: BLE001
+            logger.warning(
+                f"write_hf_assets failed for {ckpt_dir}: {e} -- checkpoint "
+                f"weights are saved; skipping HF assets."
+            )
+        try:
```
tiankongdeguiji: Keep prompt export out of checkpoint saving. A training checkpoint should contain model, optimizer, and optional scalar contract metadata; it should not copy serving artifacts through the model. Build the tokenizer and scripted front-end from pipeline.config during export.

tzrec/main.py
```diff
@@ -1077,6 +1110,53 @@ def export(
...
+                f"was not written by an HF-backed model."
+            )
+        if assets:
+            logger.warning(f"HF export ignores asset_files: {assets}.")
+        if is_rank_zero:
+            from tzrec.utils.hf_export_util import dcp_to_hf
+
+            dcp_to_hf(checkpoint_path, export_dir)
+            # Carry the prompt contract saved alongside the checkpoint.
+            copy_prompt_assets(checkpoint_path, export_dir)
```
tiankongdeguiji: Generate the actual serving contract during export. Export the resized HF model, extended tokenizer, scripted GenRec front-end, and embedded identity digests from the compiled pipeline. Do not copy prompt JSON files from a training checkpoint.

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+            lm_parameter_dtype: dtype of the LM parameters.
+
+        Returns:
+            A randomly initialized backbone with the requested parameter dtype.
+        """
+        config = AutoConfig.from_pretrained(hf_model_name_or_path)
+        model = AutoModelForCausalLM.from_config(config)
+        return model.to(_PARAM_DTYPE[lm_parameter_dtype])
+
+    def _build_projections(self) -> None:
```
tiankongdeguiji: Rename this mutating helper to init_projections(). It initializes self.projections and self._slot_projections rather than returning a built value, so the init_* lifecycle naming is clearer and consistent with init_input().

tzrec/prompt/persist_test.py
```diff
@@ -0,0 +1,223 @@
+# Copyright (c) 2026, Alibaba Group;
```
tiankongdeguiji, line 1: Rename this module to types.py. It defines the complete compiled-prompt type system, including SID space, segments, plans, and CompiledPrompt; plan.py under-describes its contents.

tzrec/prompt/assembler.py
```diff
@@ -0,0 +1,315 @@
...
+        if np.any(by_level < self._flat_lo) or np.any(by_level >= self._flat_hi):
+            raise ValueError(
+                f"prompt slot [{name}]: SID values must already carry their "
+                f"level offset, so level l lies in "
+                f"[level_offsets[l], level_offsets[l] + codebook[l]). Read the "
+                f"offset_codebook column, not codebook or origin_codebook."
+            )
+        return values.astype(np.int64, copy=False) + self._sid_space.base_vocab_size
+
+    def _emit_sample(
```
tiankongdeguiji: Remove the per-sample emission path. Building Python lists and slicing each row separately makes assembly scale with batch size in Python and cannot be TorchScripted. The compiled segment walk should operate directly on flat values and jagged lengths.

tzrec/prompt/assembler.py
```diff
@@ -0,0 +1,315 @@
...
+                    range(
+                        sample_start + len(out),
+                        sample_start + len(out) + width,
+                    )
+                )
+                projected_occurrence_index += 1
+                out.extend([self._sid_space.sentinel_token_id] * width)
+        return projected_occurrence_index
+
+    def assemble(
```
tiankongdeguiji: Remove this row-oriented assemble() API. Keep one tensor-based batch implementation shared by the training collator and exported serving front-end; a separate Python/NumPy assembly path would create both a performance bottleneck and a second prompting implementation.

tzrec/prompt/assembler.py
```diff
@@ -0,0 +1,315 @@
...
+            for position in occurrence_positions
+        ]
+        return AssembledPrompt(
+            input_ids=np.asarray(jagged_token_ids, dtype=np.int64),
+            cu_seqlens=np.asarray(cu_seqlens, dtype=np.int64),
+            hole_positions=np.asarray(holes, dtype=np.int64),
+            response_lengths=np.asarray(response_lengths, dtype=np.int64),
+        )
+
+    def assemble_batch(
```
tiankongdeguiji: Use forward() instead of assemble_batch(). This assembler should have no single-sample path and always consumes jagged batch tensors, so the _batch suffix is redundant. Making it the standard module entry point also lets training and TorchScript serving call the same implementation directly.

tzrec/prompt/assembler.py
```diff
@@ -0,0 +1,315 @@
...
+
+        jagged_token_ids: List[int] = []
+        response_lengths: List[int] = []
+        holes_by_occurrence = [
+            []
+            for seg in self._prompt_plan.segments + self._response_segments
+            if isinstance(seg, SlotSeg) and seg.fill is FillMode.PROJECTED
+        ]
+        cu_seqlens = [0]
+        for sample_index in range(batch_size):
```
tiankongdeguiji: Do not loop over samples. Compute per-segment lengths, row offsets, and segment destinations with cumsum and repeat_interleave, then fill the flat token buffer with index_copy. A static loop over compiled segments is acceptable; a data-dependent loop over the batch is not.

tzrec/models/prompt_generative_qwen.py
```diff
@@ -0,0 +1,267 @@
...
+            batch: carries the packed prompt in ``additional_infos``.
+
+        Returns:
+            The loss when training, the decoded SIDs otherwise.
+        """
+        # outside the leaf on both paths, so the pipeline can prefetch it
+        embeds = self.build_input(batch)
+        if self.is_inference:
+            return {self._generated_sids_key: _fx_wrapped_generate(self, embeds, batch)}
+        return _fx_wrapped_loss(self, embeds, batch)
```
tiankongdeguiji: Return response-window logits instead of computing loss in predict(). This preserves the suffix-only memory bound while allowing TorchEasyRec's loss() stage to invoke the backbone-owned causal-LM loss.

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+        """Surface the CE already computed in ``predict``.
+
+        Args:
+            predictions: what ``predict`` returned.
+            batch: the batch, unused.
+
+        Returns:
+            The named loss.
+        """
+        return {"ce_loss": predictions["loss"]}
```
tiankongdeguiji: Compute CE here with self.lm.loss_function. Consume the response-window logits from predict() and labels derived from the batch instead of forwarding a loss scalar as a prediction.

tzrec/models/prompt_generative_model.py
```diff
@@ -0,0 +1,269 @@
...
+        losses: Optional[Dict[str, torch.Tensor]] = None,
+    ) -> None:
+        """Update the mean-CE metric with this batch's loss.
+
+        Args:
+            predictions: what ``predict`` returned.
+            batch: the batch, unused.
+            losses: the named losses, unused.
+        """
+        self._metric_modules["ce_loss"].update(predictions["loss"].detach())
```
tiankongdeguiji: Consume the loss supplied by the evaluation loop. main already passes the computed losses into update_metric, so update ce_loss from losses["ce_loss"] rather than carrying the same scalar through predictions.

tzrec/models/genrec_causal_lm_model.py
```diff
@@ -0,0 +1,267 @@
...
+        window only: a full (batch, length, vocab) upcast does not fit.
+
+        Args:
+            embeds: the assembled prompt embeddings, packed.
+            batch: carries the row boundaries and the collator's width.
+
+        Returns:
+            The loss.
+        """
+        padded, mask, labels = self._left_pad_packed_inputs(
```
tiankongdeguiji, line 146: Follow up with packed-varlen execution in the next PR. This dense-padding adapter is acceptable for initial HuggingFace compatibility, but it expands flat embeddings into (B, T, D) and wastes memory and attention compute on padding. The next PR should run Qwen directly from cu_seqlens and gather response positions from packed hidden states for loss computation.

tiankongdeguiji (changes requested):

tzrec/datasets/dataset.py
```diff
@@ -121,8 +135,25 @@ def __init__(
             else None
         )
 
+        parser_features = features
```
tiankongdeguiji: `parser_features` should not exist — the answer is a label, not a feature

`tzrec/datasets/dataset.py:138-156` makes the parsed feature set mode-dependent: at
`PREDICT` it walks the compiled plan, subtracts the response slots' members from the
prompt slots', and hands `DataParser` a reduced list. It is there because the
supervised answer is declared in `feature_configs`, and a feature is an input column
the parser unconditionally requires — while at predict there is no answer column. The
filter is a patch undoing a declaration that shouldn't be made.

Three costs:

1. **One fact, two derivations.** `PromptAssembler(include_response=mode != PREDICT)`
   (`prompt/assembler.py:79-85`) decides what the assembler *reads*; the set arithmetic
   in the dataset decides what the parser *produces*. Nothing checks the pair, and they
   are computed by different code. Drift is a `KeyError` in `assemble_batch`, or a
   reader demanding a column that isn't there.
2. **Prompt semantics in the generic dataset.** `BaseDataset` now imports `SlotSeg` and
   walks `segments` / `response_segments`. It is the only place in `dataset.py` that
   knows what a model's target looks like.
3. **It doesn't stop here.** Under `FG_DAG` the FG graph `DataParser` builds
   (`data_parser.py:177`) now differs between train and predict, and everything derived
   from the full feature list still advertises the answer as an input:
   `create_fg_json(features)` in `export_util.py`, and the `feature_configs` rewritten
   into the exported `pipeline.config`. RFC 0002 §5.1 has serving run FG from that
   published `fg_json`, so a serving host would be told to produce an `item:answer`
   input that no online request can supply — a third place the same subtraction has to
   be re-derived.

**The response feature contributes nothing to the compiled plan.** Of what a `SlotSeg`
derives (`prompt/compile.py:319-336`):

| derived | for a response slot | source |
| --- | --- | --- |
| `width` | `STATIC(num_levels)` | `sid_space`, never the feature (`compile.py:93-94`) |
| `fill` | INLINE | forced — PROJECTED is rejected at `compile.py:284-288` |
| `group_type` | JAGGED_SEQUENCE | forced — a scalar member is DEEP → PROJECTED → rejected |
| shift | `base_vocab_size` | applied unconditionally in `assembler.py:_inline_tokens` |

The feature declaration's only remaining job is to satisfy "every `feature_names` entry
exists in `feature_configs`" — which then forces the filter that cancels the parse it
caused.

**And the column is already declared twice.** `prompt_generative_qwen_mock.config` has
both `label_fields: "answer"` (:31) and `sequence_raw_feature { feature_name: "answer" }`
(:42). `DataParser.parse` writes features first (`data_parser.py:289-321`, keys
`answer.values` / `answer.lengths`) and labels second (`data_parser.py:221-250`, **the
same keys**), so the label pass already overwrites the feature pass. At train the
assembler is reading label-parsed tensors; the feature pass produces a dense
`(total, value_dim)` float array that is then silently replaced by flat int64.

#### Proposal: a `response` placeholder resolves to a `data_config.label_fields` column

`compile_prompt(cfg, features, label_fields, ...)` keeps resolving body placeholders
against features, and resolves response placeholders against `label_fields`, building
those `SlotSeg`s from the constants in the table above. Everything else already exists:

- `DataParser(labels=... if mode != PREDICT else None)` is the mode gate, in one place.
- Labels already emit `{name}.values` / `{name}.lengths` for `list<int>` columns into
  the same `output_data` the assembler consumes, so **the assembler needs no change**
  (no `key_lengths` → `per_row = lengths` → reshape `(-1, num_levels)` → band-check).
- `_selected_input_names` already excludes `label_fields` at predict.

Deleted: `dataset.py:138-156` and its `SlotSeg` import; the duplicate `feature_configs`
block; `_slot_width`'s `answer_levels` parameter; the response-is-PROJECTED check at
`compile.py:284-288`, i.e. RFC 0001 §5.3 rule 6, which becomes structurally impossible
rather than enforced. Added: one check that every response placeholder names a declared
label field.

Precedent in this branch's own history: `b373e9e5` *"[refactor] generative-rec LM: answer
is a `data_config.label_field`, not a feature"* made exactly this move in the pre-prompt
stack, for the same reason — it "lets the label be absent at inference without the
EmbeddingGroup requiring it". The prompt rewrite reverted it.

**On a future text response:** tokenize it with the extended tokenizer `compile_prompt`
already builds and persists (covered by `vocab_hash`) — not through FG. FG is the
feature extraction layer; routing a label through it means RFC 0001 §3.4's `vocab_file`
injection, i.e. a second tokenizer instance, configured by path, that can disagree with
the vocabulary `lm_head` generates into. Using the same in-process `Tokenizer` the
compiler extended makes that disagreement unrepresentable. The label route is not a
trade against text responses; it is the right shape for them too.

**Related, same code:** the answer-width check at `compile.py:515-526` cannot fire — a
response slot's width is `STATIC(num_levels)` by construction, so the comparison is
tautological. The check that is missing is on the data: `_inline_tokens` only asserts
`values.size % num_levels == 0` (`assembler.py:110`), so a label row carrying two items
passes, and `logits_suffix_len` — derived from the declared width — then opens a
supervised window shorter than what was emitted, dropping part of the loss silently.
With the answer as a label, the exact width is known at compile, so this becomes
`values.size == num_levels`.

tiankongdeguiji (changes requested):

WhiteSwan1:
@tiankongdeguiji  Thank you for the detailed review. We went through both review rounds and reworked the branch around the framework boundaries you called out.

## Changes made

- Renamed the stack around GenRec rather than prompt rendering: `BaseGenrecModel`, `GenrecCausalLMModel`, and `prompt/types.py`. The mutating setup methods are now `init_backbone()` and `init_projections()`. Backbone construction also validates the configuration, embedding, loss, and `forward(logits_to_keep=...)` interfaces used by this implementation.
- Restored the TorchEasyRec model contract. `predict()` now returns the bounded response-window logits and labels, `loss()` invokes `self.lm.loss_function`, and `update_metric()` consumes `losses["ce_loss"]`. Training explicitly disables the KV cache and calls the causal LM through its public forward interface.
- Made the answer label-only. Response placeholders resolve exclusively against `data_config.label_fields`; `parser_features` and the dataset's prompt-plan walk were removed, as were the redundant answer entries in `feature_configs`. A response must name one label field, and each row must match the compiled SID width, so a wider answer can no longer be partially supervised.
- Replaced the row-oriented assembler with one batch-wide `forward()` path. It consumes flat values and jagged lengths, computes row and segment offsets in bulk, and fills the packed output while looping only over the static compiled segment list. There is no data-dependent Python loop over samples.
- Precomputed SID bands and capped beam widths during model initialization instead of rebuilding and revalidating them for every inference call.
- Removed `export_format` and dispatch export from the configured model class. The model-level asset hook, persisted `PromptPlan`, `sid_space.json`, and the standalone hash artifact were removed. For the current converter, the checkpoint still retains its HF configuration plus FQN and prompt-digest metadata. Export verifies that the checkpoint vocabulary matches the freshly compiled prompt and regenerates the extended tokenizer from `pipeline.config`.
- The current direct DCP-to-HF path now rejects projected slots before creating output. It therefore cannot silently publish a backbone-only artifact while dropping sparse embeddings or projection parameters.

## Deliberate scope and follow-ups

### `init_backbone()` ownership

We adopted the requested lifecycle shape: the method assigns `self.lm` and returns `None`. We kept it on `BaseGenrecModel` because the base class owns all of the related LM lifecycle: vocabulary resize, projection sizing from the LM hidden size, input embedding lookup, loss, cold-start pretrained loading, and checkpoint/export access. Moving only construction to `GenrecCausalLMModel` would split ownership without removing the base class's dependency on the backbone.

### Full-model and scripted export

The remaining export-related work is not considered in this PR. The current converter exports the HF backbone and extended tokenizer, but no prompt front-end or assembler; projected prompts fail explicitly. Restoring the complete TorchEasyRec model, exporting a scripted GenRec front-end with sparse embeddings and projection state, embedding the contract digests in the serving artifacts, and generating the HF assets from the restored model are coupled parts of one serving contract. We will address them together in a subsequent export change rather than land a partial front-end here.

### Shared scripted assembler

The per-sample assembly bottleneck and row API are gone, but the current training assembler remains a NumPy implementation in dataloader workers. Tensorizing that training path alone would not satisfy the requested sharing: the exported front-end must invoke the same module, with a defined serving contract for its outputs. Since that export path is deferred, the Tensor/TorchScript conversion is also not considered in this PR; it will be implemented together with the full-model and scripted-front-end export described above.

### Packed-varlen execution

Packed-varlen Qwen execution remains the next-PR optimization requested in the review. This PR keeps the dense HuggingFace compatibility adapter, so this optimization is not considered in this PR for now.

The model identity is no longer Qwen-specific, but this PR only claims the backbone families covered by CI. Initialization rejects models missing the stable interfaces listed above. Broader validation and adaptation of cache protocols across causal-LM families is not included in this change.

WhiteSwan1:
Thank you for the detailed review. We checked all 13 inline findings against the current supported execution and restore paths. We made two follow-up corrections and deliberately left the remaining performance, compatibility, and test-only suggestions unchanged for the reasons below.

## Changes made

- Checkpoint asset failures are no longer downgraded to warnings. If rank 0 fails to write the HF assets, the error is broadcast to every rank and the save fails before writing dataloader state, checkpoint metadata, `_SUCCESS`, or scheduling pruning. Already-written checkpoint shards may remain for diagnosis, but the job stops and this save cannot prune an older healthy checkpoint. A two-rank Gloo regression test verifies that both ranks receive the failure without hanging and that none of the completion artifacts are written.
- The `PromptConfig.response` comment now matches the actual contract: the response must be declared in every mode, while inference assembly omits it from the rendered prompt.

## Prompt plan mismatch policy

We are keeping `plan_hash` mismatches as warnings. The routing-swap scenario is technically possible, but it requires the caller to restore a PROJECTED checkpoint with an explicitly changed prompt configuration and then ignore the existing mismatch warning; it cannot arise from the normal same-config train, continue, evaluate, or predict paths. A global fatal check would also reject intentional prompt and topology changes supported by `fine_tune_checkpoint` and partial restore.

A projection-only digest would not be a complete weight contract: it would still omit feature expressions and bucket semantics, resolved embedding dimensions, the LM architecture, DCP partial-load results, and implementation-version semantics. The current primary configuration is INLINE-only, and HF export already refuses PROJECTED prompts. If PROJECTED serving later requires strict reproducibility, the contract should distinguish strict continue/evaluate/predict/serving restores from explicitly permissive fine-tuning rather than make every plan change fatal.

## Decode memory

The full-vocabulary FP32 normalization is a real peak-memory cost at large `batch_size * beam_width`, but it does not produce incorrect scores when it completes. Full-vocabulary normalization is required to compare candidates from different parent beams; replacing it with band-only normalization would change beam ranking. A meaningful memory reduction requires chunked or fused FP32 log-sum-exp, not merely reusing the full FP32 tensor. That larger decode optimization is not considered in this PR.

## Dependencies and interface checks

- Leaving `transformers` unpinned is deliberate. The supported Qwen paths have been exercised with Transformers 5.16.1 on H20, while `tokenizers` is a required dependency of Transformers rather than an optional package that can be absent from a normal installation.
- We are not adding a construction-time Cache probe. The supported Qwen2/Qwen3 implementations provide the modern Cache protocol upstream, while reliably probing an arbitrary model's returned cache would require a dummy forward and would duplicate upstream behavior.
- We are not changing `check_prompt_assets(None, ...)`. A GenRec model without a compiled prompt is already rejected during construction; restoring a checkpoint into a different model class falls under the framework's existing partial-restore behavior.
- The supported HF conversion entry point remains deliberately dist-free. The hypothetical rank-zero-only call from an already initialized process group is not a supported export path and does not warrant a new process-group fallback in this change.

## Test-only and minor performance suggestions

The inference-composition, dataloader, assembler-assertion, and individual export-guard comments request additional coverage without identifying a concrete production failure. We are not expanding test-only coverage in this follow-up. We are also leaving the proposed in-place `index_copy_` change out because it is a small PROJECTED-only optimization without target-environment FX/DMP evidence.

tiankongdeguiji (approved):

PR #638, opened 2026-08-21
[feat] tombstone evicted dynamic embedding keys in delta dump

eric-gecheng:
Evicted dynamicemb keys were silently skipped at dump time, so their stale FeatureStore rows kept serving forever and the processor could never reclaim the memory. The dump now drains pop_evicted_keys after the tracker pass (capturing flush-induced evictions in the same window), subtracts keys already republished as real rows, and uploads the rest as all-zero tombstones through the existing chunk/quant path; the polling processor detects the zeros and deletes the key from NvEmbeddings. The new dump_evicted_tombstones switch (opt-in, default false) gates the pass and auto-arms dynamicemb RETAIN_KEY recording for exactly those runs, degrading to a one-time warning on old dynamicemb builds. Retained keys live in per-rank GPU memory until the next dump drains them, so the switch stays opt-in rather than a silent default for existing delta-dump jobs.

### DIFF
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -907,36 +907,43 @@ def train_and_evaluate(
         with open(os.path.join(pipeline_config.model_dir, "version"), "w") as f:
             f.write(tzrec_version + "\n")
 
-    if delta_embedding_dumper is not None:
-        delta_embedding_dumper.start()
-    # when slice batch by sample cost, data on all workers may not be balanced
-    check_all_workers_data_status = data_config.HasField("batch_cost_size")
-    _train_and_evaluate(
-        model,
-        optimizer,
-        train_dataloader,
-        eval_dataloader,
-        [sparse_lr, dense_lr, *part_lrs],
-        pipeline_config.model_dir,
-        train_config=train_config,
-        eval_config=pipeline_config.eval_config,
-        ckpt_manager=ckpt_manager,
-        skip_steps=skip_steps,
-        ckpt_path=ckpt_path,
-        check_all_workers_data_status=check_all_workers_data_status,
-        ignore_restore_optimizer=ignore_restore_optimizer,
-        dataloader_state=dataloader_state,
-        delta_embedding_dumper=delta_embedding_dumper,
-        pipeline_config_path=os.path.join(pipeline_config.model_dir, "pipeline.config"),
-        dense_ema=dense_ema,
-        export_config=pipeline_config.export_config,
-    )
-    # Drain background uploads only after training succeeds. A training failure
-    # terminates the whole job (torchrun tears down every rank) and pending
-    # in-memory deltas are intentionally abandoned: the restarted run re-dumps
-    # from the latest checkpoint, so there is nothing to roll back or undo.
-    if delta_embedding_dumper is not None:
-        delta_embedding_dumper.close()
+    try:
+        if delta_embedding_dumper is not None:
+            delta_embedding_dumper.start()
+        # when slice batch by sample cost, data on all workers may not be balanced
+        check_all_workers_data_status = data_config.HasField("batch_cost_size")
+        _train_and_evaluate(
+            model,
+            optimizer,
+            train_dataloader,
+            eval_dataloader,
+            [sparse_lr, dense_lr, *part_lrs],
+            pipeline_config.model_dir,
+            train_config=train_config,
+            eval_config=pipeline_config.eval_config,
+            ckpt_manager=ckpt_manager,
+            skip_steps=skip_steps,
+            ckpt_path=ckpt_path,
+            check_all_workers_data_status=check_all_workers_data_status,
+            ignore_restore_optimizer=ignore_restore_optimizer,
+            dataloader_state=dataloader_state,
+            delta_embedding_dumper=delta_embedding_dumper,
+            pipeline_config_path=os.path.join(
+                pipeline_config.model_dir, "pipeline.config"
+            ),
+            dense_ema=dense_ema,
+            export_config=pipeline_config.export_config,
+        )
+        # Drain background uploads only after training succeeds. A training failure
+        # terminates the whole job (torchrun tears down every rank) and pending
+        # in-memory deltas are intentionally abandoned: the restarted run re-dumps
+        # from the latest checkpoint, so there is nothing to roll back or undo.
+        if delta_embedding_dumper is not None:
+            delta_embedding_dumper.close()
+    finally:
+        # Evicted-key retention is process-global; disarm so anything running
+        # after training in this process never sees it left armed.
+        dynamicemb_util.set_auto_retain_evicted_keys(False)
     if is_local_rank_zero:
         logger.info("Train and Evaluate Finished.")
 
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -907,43 +907,36 @@ def train_and_evaluate(
         with open(os.path.join(pipeline_config.model_dir, "version"), "w") as f:
             f.write(tzrec_version + "\n")
 
-    try:
-        if delta_embedding_dumper is not None:
-            delta_embedding_dumper.start()
-        # when slice batch by sample cost, data on all workers may not be balanced
-        check_all_workers_data_status = data_config.HasField("batch_cost_size")
-        _train_and_evaluate(
-            model,
-            optimizer,
-            train_dataloader,
-            eval_dataloader,
-            [sparse_lr, dense_lr, *part_lrs],
-            pipeline_config.model_dir,
-            train_config=train_config,
-            eval_config=pipeline_config.eval_config,
-            ckpt_manager=ckpt_manager,
-            skip_steps=skip_steps,
-            ckpt_path=ckpt_path,
-            check_all_workers_data_status=check_all_workers_data_status,
-            ignore_restore_optimizer=ignore_restore_optimizer,
-            dataloader_state=dataloader_state,
-            delta_embedding_dumper=delta_embedding_dumper,
-            pipeline_config_path=os.path.join(
-                pipeline_config.model_dir, "pipeline.config"
-            ),
-            dense_ema=dense_ema,
-            export_config=pipeline_config.export_config,
-        )
-        # Drain background uploads only after training succeeds. A training failure
-        # terminates the whole job (torchrun tears down every rank) and pending
-        # in-memory deltas are intentionally abandoned: the restarted run re-dumps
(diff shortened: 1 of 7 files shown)


tzrec/utils/delta_embedding_dump.py
```diff
@@ -1130,6 +1205,94 @@ def _lookup_dynamic_embeddings(
...
+            emb_dim = dynamic_module._dynamicemb_options[table_id].dim
+            num_rows += self._append_table_chunk(
+                table_chunks,
+                global_step=global_step,
+                feature_name=_feature_name(
+                    self._tracker.fqn_to_feature_names.get(fqn, [])
+                ),
+                table_fqn=fqn,
+                key_ids=evicted,
+                embeddings=torch.zeros((evicted.numel(), emb_dim), dtype=torch.float32),
```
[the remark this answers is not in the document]
eric-gecheng: unneccessary

tzrec/protos/train.proto
```diff
@@ -108,6 +108,13 @@ message DeltaEmbeddingDumpConfig {
     // even embedding_dim.
     optional DeltaEmbeddingQuantType quant_type = 6
         [default = DELTA_EMBEDDING_QUANT_NONE];
+    // Publish all-zero tombstone rows for dynamicemb keys evicted since the
+    // last dump; the polling processor detects them and deletes the key from
+    // NvEmbeddings, reclaiming its memory (FeatureStore has no delete). While
+    // enabled, delta dump mode also arms dynamicemb's evicted-key recording
+    // (evicted_item_mode=RETAIN_KEY) automatically. Requires a dynamicemb
+    // build with EvictedItemMode; older builds degrade to a warning.
+    optional bool dump_evicted_tombstones = 7 [default = true];
```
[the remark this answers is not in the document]
eric-gecheng: Eviction is inherent to dynamicemb and downstream has no delete path, so without tombstones every evicted key serves its stale row forever — deletion semantics are required by default, not opt-in. The host-side OOM is fixed in 60a1844 (bounded chunks sharing one chunk-sized zero buffer; the key-only GPU buffer stays documented on the field).

tzrec/utils/delta_embedding_dump.py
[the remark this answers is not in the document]
eric-gecheng: Fixed in 60a1844: tombstones are now emitted in 65536-row chunks whose embeddings all alias one chunk-sized zero buffer — _append_table_chunk wraps the row-slice views zero-copy, so the Arrow chunks, concat_tables, and the uploader's batch slicing all share that single buffer. Host memory for tombstone zeros is capped at ~32 MB (dim 128) per table per dump, independent of eviction volume.

tzrec/main.py
```diff
@@ -689,6 +694,12 @@ def train_and_evaluate(
     is_local_rank_zero = int(os.environ.get("LOCAL_RANK", 0)) == 0
     acc_utils.allow_tf32(train_config)
     enable_delta_embedding_dump = train_config.HasField("delta_embedding_dump_config")
+    # Arm evicted-key retention before feature/model building so the dump's
+    # pop_evicted_keys drain has a buffer to consume on every dynamicemb table.
+    dynamicemb_util.set_auto_retain_evicted_keys(
```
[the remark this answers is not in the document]
eric-gecheng: Fixed in f9bbbda: the training body is wrapped in try/finally and retention is disarmed in the finally, so the process-global flag resets whether training succeeds or fails. close() stays inside the try to preserve the existing "drain uploads only on success" semantics.

eric-gecheng:
@copilot resolve the merge conflicts in this pull request

tzrec/main.py
```diff
@@ -896,36 +907,43 @@ def train_and_evaluate(
...
-        dense_ema=dense_ema,
-        export_config=pipeline_config.export_config,
-    )
-    # Drain background uploads only after training succeeds. A training failure
-    # terminates the whole job (torchrun tears down every rank) and pending
-    # in-memory deltas are intentionally abandoned: the restarted run re-dumps
-    # from the latest checkpoint, so there is nothing to roll back or undo.
-    if delta_embedding_dumper is not None:
-        delta_embedding_dumper.close()
+    try:
```
tiankongdeguiji: `train_and_evaluate()` runs in a one-shot subprocess, so resetting this process-global flag has no practical effect—the process exits after training, and the reset does not affect already-built tables. Please remove the `try/finally` and keep the existing training flow unchanged; tests can reset the flag in cleanup.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -1130,6 +1208,105 @@ def _lookup_dynamic_embeddings(
...
+        self,
+        table_chunks: List[pa.Table],
+        global_step: int,
+        dynamic_modules: Dict[str, nn.Module],
+        published_key_ids: Dict[str, torch.Tensor],
+        flushed_module_ids: Set[int],
+    ) -> int:
+        """Append zero-row tombstones for dynamicemb keys evicted since the last dump.
+
+        The authoritative eviction source is ``pop_evicted_keys`` (read-and-clear
```
tiankongdeguiji: simplify comments

tzrec/protos/train.proto
```diff
@@ -108,6 +108,18 @@ message DeltaEmbeddingDumpConfig {
     // even embedding_dim.
     optional DeltaEmbeddingQuantType quant_type = 6
         [default = DELTA_EMBEDDING_QUANT_NONE];
+    // Publish all-zero tombstone rows for dynamicemb keys evicted since the
```
tiankongdeguiji, line 111: simplify comments

tzrec/utils/delta_embedding_dump.py
tiankongdeguiji, line 786: simplify comments

tzrec/utils/delta_embedding_dump.py
```diff
@@ -983,6 +1054,13 @@ def _append_model_delta_rows(
                 dynamic_modules=dynamic_modules,
                 flushed_module_ids=flushed_module_ids,
             )
+            if published_dynamic_key_ids is not None and fqn in dynamic_modules:
```
tiankongdeguiji: For DynamicEmb, `key_ids` is already copied D2H here and then copied again in `_append_table_chunk()`. Also, since `get_unique()` returns one entry per FQN and the map starts empty, the `torch.cat()` branch is unnecessary. Please materialize `key_ids_cpu` once and reuse it for both tracking and table construction.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -965,12 +1016,32 @@ def _append_model_delta_rows(
...
+        """
         num_rows = 0
         # A dynamic module hosting multiple tables is shared across their FQN
         # keys; flush() flushes the whole module, so track which
         # modules were already flushed this dump and skip the redundant repeats.
-        flushed_module_ids: Set[int] = set()
+        if flushed_module_ids is None:
+            flushed_module_ids = set()
         for fqn, unique_rows in self._tracker.get_unique(_CONSUMER).items():
             ids = unique_rows.ids
```
tiankongdeguiji: Could we merge pop_evicted_keys() with unique_rows.ids and run a single lookup after flush()? Found keys can publish their current embeddings, while missing keys originating from the evicted set can publish zero tombstones. This handles both evict-then-reinsert and insert-then-evict correctly

tiankongdeguiji (changes requested):

tzrec/utils/delta_embedding_dump.py
```diff
@@ -1093,42 +1162,193 @@ def _lookup_zch_embeddings(
...
+            batch_ids = ids[start : start + _FIND_BATCH_ROWS]
+            gpu_ids = batch_ids.to(device=device)
+            table_ids = torch.full_like(gpu_ids, table_id, dtype=torch.int64)
+            _, _, _, _, _, founds, _, values = dynamic_module.tables.find(
+                gpu_ids, table_ids, CopyMode.EMBEDDING
+            )
+            founds = founds.to(dtype=torch.bool)
+            missing = ~founds
+            if evicted_mask is not None:
+                batch_evicted = evicted_mask[start : start + _FIND_BATCH_ROWS]
```
tiankongdeguiji, line 1274: evicted_mask is on CPU, while missing returned by find() is on CUDA, so missing & batch_evicted will fail with a device mismatch. Why isn’t this case covered by a test?

tzrec/utils/delta_embedding_dump.py
```diff
@@ -1093,42 +1162,193 @@ def _lookup_zch_embeddings(
...
+        if table_fqn in self._warned_no_retain_tables:
+            return
+        self._warned_no_retain_tables.add(table_fqn)
+        logger.warning(
+            "Delta embedding dump cannot tombstone evicted keys for table "
+            f"{table_fqn}: {reason}; stale FeatureStore rows for its evicted "
+            "keys will never be reclaimed."
+        )
+
+    def _append_tombstone_chunks(
```
tiankongdeguiji, line 1315: Treating every evicted key as an all-zero embedding is unsafe. DynamicEmb supports configurable [`[eval_initializer_args](https://torcheasyrec.readthedocs.io/zh-cn/latest/feature/dynamicemb.html)`](https://torcheasyrec.readthedocs.io/zh-cn/latest/feature/dynamicemb.html), so the evaluation value of an evicted/missing key is not necessarily zero, while a valid live embedding may legitimately be all zero. Please merge the popped evicted IDs with `unique_rows.ids` and resolve all of them through the same DynamicEmb lookup path instead of fabricating zero values; otherwise a valid key may be misinterpreted and deleted.

tzrec/utils/delta_embedding_dump.py
```diff
@@ -1093,42 +1162,193 @@ def _lookup_zch_embeddings(
...
+            emb_dim: The table's embedding dimension.
+
+        Returns:
+            Number of tombstone rows appended.
+        """
+        if tombstone_ids.numel() == 0:
+            return 0
+        feature_name = _feature_name(self._tracker.fqn_to_feature_names.get(fqn, []))
+        chunk_rows = min(_TOMBSTONE_CHUNK_ROWS, tombstone_ids.numel())
+        zero_rows = torch.zeros((chunk_rows, emb_dim), dtype=torch.float32)
```
tiankongdeguiji: INT8 is supported or not?

tiankongdeguiji (changes requested):

tiankongdeguiji (approved):

PR #648, opened 2026-08-27
[feat] support weighted id feature not using weight in model

tiankongdeguiji:
## Motivation

`IdFeature.weighted` tells fg that the input carries `id:weight` pairs (a `k1:v1\x1dk2:v2` string or a `map<*,float>` column). Some models only need the ids from such an input and do not want the weight to take part in the embedding pooling, but there was no way to say so: the weight was always parsed, written into the data group's `KeyedJaggedTensor` and shipped through `input_dist`.

While adding that knob it turned out that training and serving did not agree on those weights. `EmbeddingBagCollection` was constructed with the default `is_weighted=False`, and training and export treat that flag differently:

- the sharded training lookup reads `features.weights_or_none()` unconditionally (`torchrec/distributed/batched_embedding_kernel.py`, `BaseBatchedEmbeddingBag.forward`), so DMP training **did** apply the weights;
- both export modules gate them on the flag and therefore **dropped** them — the quantized `QuantEmbeddingBagCollection` (`torchrec/quant/embedding_modules.py:554`) that `QUANT_EMB=1` produces, and the eager `EmbeddingBagCollection` (`torchrec/modules/embedding_modules.py:253`) that `QUANT_EMB=0` produces.

Measured on a table whose row `i` is `[i, i, i, i]`, values `[1, 2]`, weights `[10, 100]` (unweighted sum = 3, weighted sum = 210):

| path | kernel | `is_weighted=False` | `is_weighted=True` |
|---|---|---|---|
| training (DMP, 1 GPU) | `SplitTableBatchedEmbeddingBagsCodegen` | 210 | 210 |
| export, `QUANT_EMB=1` | `IntNBitTableBatchedEmbeddingBagsCodegen` | 3 | 210 |
| export, `QUANT_EMB=0` | `nn.EmbeddingBag` | 3 | 210 |

## Changes

- New `IdFeature.use_weight`, **default `false`**. `weighted` keeps its meaning, "the input format carries weights, fg has to split them off the value"; `use_weight` decides whether the model consumes them. By default the weight is parsed off and dropped, so no `<feat>.weights` tensor is produced and the group's KJT stays unweighted.
- When `use_weight` is on, the `EmbeddingBagCollection` holding the feature is created with `is_weighted=True`, so the exported model pools the same way training does. `is_weighted` is a property of a whole collection, so each collection — `ebc`, `mc_ebc` and their input-tile variants — is weighted only when one of its own tables is looked up with per-sample weights. `emb_bag_config` stamps that on the table as `is_weighted` (next to the existing `trainable` and `use_dynamicemb` stamps), true when the feature's own weight is used or when it is a mean pooled table whose mean is carried by the normalized weights; `_add_embedding_bag_config` ORs it when a table is shared.
- **Mean pooling is carried by the weights.** `DataParser` normalizes the per-sample weights of a mean pooled feature to sum to one per bag, and `BaseFeature.emb_bag_config` declares its table with sum pooling. Pooling is then a plain weighted sum in every kernel. `is_weighted` is collection-level, so the unweighted features of the group are normalized the same way — their all-ones weights become `1/L` — and their tables are converted too. Since that is a data-group-level fact rather than a per-feature one, `create_features` marks every feature whose data group contains a `use_weight` feature (`in_weighted_group`), which is safe there because data groups are only ever assigned in that function.
- Raw weights stay in the parsed tensor dict; normalization happens in `to_batch`. So serving passes the weights it always did, and `dump_parsed_inputs` still round-trips to the original `id:weight` input.
- `dynamicemb` features cannot share a data group with a weighted feature — see below — and `emb_bag_config` raises a config error for that.
- Parsing a map input in `FG_NONE` mode used to extract the weight regardless of the `weighted` flag, leaving an unused weight tensor in the parsed inputs. The map branch is now gated on `weighted` like the string branch.

### Why normalize rather than fix the divisor

A weighted mean has no single definition in this stack, measured on the same input:

| path | who pools | weighted mean |
|---|---|---|
| training | torchrec RW callback | `Σ(wᵢ·eᵢ) / Σwᵢ` = 1.909 |
| export, `QUANT_EMB=1` | fbgemm nbit TBE | `Σ(wᵢ·eᵢ) / L` = 105 |
| export, `QUANT_EMB=0` | `nn.EmbeddingBag` | `NotImplementedError` |

Every fbgemm kernel normalizes by the bag length and ignores the weights in the divisor — nbit inference (`embedding_forward_quantized_split_nbit_kernel_template.cu`, and `RefImplementations.cc`'s `1.f / len`) and the training kernel (`embedding_forward_split_kernel_template.cu`'s `inv_L = 1.0 / L`) alike. Training escapes that only because `pooling_type_to_pooling_mode` forces the TBE to `SUM` under RW/TWRW sharding ("Mean pooling is not supported in TBE for TWRW/RW sharding") and torchrec normalizes afterwards in `_create_mean_pooling_divisor`, which switches the divisor to `Σw`.

Since `Σ(wᵢ·eᵢ)/Σwᵢ ≡ Σ(w'ᵢ·eᵢ)` with `w'ᵢ = wᵢ/Σw`, normalizing the weights removes the divisor from the problem entirely. Sum pooling has nothing to disagree about, and it is shard-additive, so the result is the same in eager, quantized and sharded lookups without relying on either divisor convention.

### Behavior change

None for a config that does not set `use_weight: true` — the default keeps today's serving numerics and stops shipping a weight tensor the exported model never used. Opting in makes training and serving agree on `Σ(w·e)/Σw`.

Note that the previous behaviour was already inconsistent: training applied the weights, serving did not. There is no default that preserves both. `false` preserves serving; a config that was relying on the weights being applied **in training** needs `use_weight: true`.

### dynamicemb

dynamicemb does not support per-sample weights for pooling, and it repurposes the KJT weights channel to carry LFU frequency counters (`dynamicemb/shard/embedding.py` sets `features._weights = frequency_counters.float()`, and `dynamicemb/input_dist.py::_determine_output_weights` preserves them on that assumption). `BatchedDynamicEmbeddingBag` has no `forward` override, so torchrec hands it `per_sample_weights=features.weights_or_none()`; `BatchedDynamicEmbeddingTablesV2.forward` passes that straight to `prefetch(frequency_counters=...)` in training and to `dynamicemb_eval_forward`'s `frequency_counters` parameter in eval, where it is cast with `.long()`. The pooling itself takes no weight term. A weighted id feature sharing a data group with a dynamicemb table would therefore get unweighted pooling *and* corrupted admission statistics — weights below 1.0 truncate to 0. This PR turns that into an explicit error.

### Out of scope

Weighted **sequence** id features remain unsupported. `SequenceSparseData` has no weights field, and on the fg side a sequence sub-feature is routed to `SequenceIdFeature`, which never reads `weighted`, has no weights slot in its output and rejects map inputs outright.

## Test Plan

- `use_weight: false` parsing for the string, list and map encodings in both `FG_NONE` and `FG_NORMAL`, and the map-input-without-`weighted` case (`tzrec/features/id_feature_test.py`, `tzrec/features/feature_test.py`).
- `DataParser` producing no `.weights` tensor and an unweighted KJT for all three fg modes; and, for a weighted group, KJT weights of `w/Σw` for the weighted mean feature, `1/L` for a plain mean feature and untouched ones for a sum feature (`tzrec/datasets/data_parser_test.py`).
- `EmbeddingGroupImpl` output compared against a hand computed weighted sum with and without `use_weight`; tables of a weighted group declared `SUM`, with the forward compared against `Σ(w·e)/Σw`; and the dynamicemb error (`tzrec/modules/embedding_test.py`).
- `python -m tzrec.features.id_feature_test`, `tzrec.features.feature_test`, `tzrec.datasets.data_parser_test`, `tzrec.modules.embedding_test` all pass. Full `python tzrec/tests/run.py` ran green (1650 tests; the only 6 failures were HSTU tests missing their gitignored KuaiRand fixtures in a fresh worktree, and all 6 pass once the fixtures are linked in).
- Both measurement tables above were taken directly with a deterministic embedding table, on `torchrec` 1.7.0 / `torch` 2.12.1+cu129, quantizing through `torchrec.inference.modules.quantize_embeddings` with the EBC nested in a parent module as it is in a real model.
- `pre-commit run -a` and `pyrefly check` clean.

### Note on integration coverage

No existing test config sets `use_weight: true`, so the weighted path is covered by unit tests only. I tried opting `multi_tower_din_zch_fg_mock.config` in, which would have exercised it through training, `QUANT_EMB=0` / `QUANT_EMB=1` export and INPUT_TILE. It made `test_multi_tower_din_zch_with_fg_train_eval_export` flaky: 1 failure in 3 runs, 2 of 8192 elements outside the `atol=1e-5` CPU-vs-GPU comparison, against 0 failures in 3 runs without it. The quantized lookup itself is bit-identical across CPU and CUDA with and without weights, so this is downstream float32 rounding at a different operating point rather than a defect — but rather than loosen a shared tolerance I left the configs untouched. Happy to add a dedicated config for weighted end-to-end coverage if you would like one.

https://claude.ai/code/session_01GRzbYSnTHNYc1he7gkGrZB

### DIFF
--- a/docs/source/feature/feature.md
+++ b/docs/source/feature/feature.md
@@ -116,6 +116,8 @@ feature_configs {
 
 - **weighted**: 是否为带权重的Id特征，输入形式为`k1:v1\x1dk2:v2`
 
+- **use_weight**: 默认值是false，仅在**weighted**=true时生效。默认情况下权重只用于解析输入，不参与模型计算，权重不会被解析和传输。设置为true时，权重会作为per_sample_weights参与Embedding pooling；**pooling**=mean时，权重会按每个样本归一化到和为1，pooling以sum的方式完成，保证训练和推理结果一致。注意同一特征组内不能有配置了**dynamicemb**的特征（dynamicemb复用了权重通道传递频次计数，不支持带权重的pooling）
+
 - **value_dim**: 默认值是0，可以设置1，value_dim=0时支持多值ID输出
 
 - **default_bucketize_value**: （可选）指定超出词表的词的编码。当配置了default_bucketize_value时，vocab_list和vocab_dict将不会预留编码给默认值和超出词表的词，用户可完全自主控制vocab_list或vocab_dict
--- a/docs/source/feature/feature.md
+++ b/docs/source/feature/feature.md
@@ -117,7 +117,7 @@ feature_configs {
 
 - **weighted**: 是否为带权重的Id特征，输入形式为`k1:v1\x1dk2:v2`
 
-- **use_weight**: 默认值是false，仅在**weighted**=true时生效。默认情况下权重只用于解析输入，不参与模型计算，权重不会被解析和传输。设置为true时，权重会作为per_sample_weights参与Embedding pooling；**pooling**=mean时，权重会按每个样本归一化到和为1，pooling以sum的方式完成，保证训练和推理结果一致。注意同一特征组内不能有配置了**dynamicemb**的特征（dynamicemb复用了权重通道传递频次计数，不支持带权重的pooling）
+- **use_weight**: 默认值是false，仅在**weighted**=true时生效。默认情况下，权重仅用于从`k:v`格式的输入中解析出id，解析后即被丢弃，不参与模型计算，也不会传输给模型。设置为true时，权重会作为per_sample_weights参与Embedding pooling；**pooling**=mean时，权重会按每个样本归一化到和为1，pooling以sum的方式完成，保证训练和推理结果一致。注意：开启后，同一**数据组（data group）**内的所有特征都会带上权重（其余特征填充为1）并走带权重的pooling kernel，会有一定的内存和计算开销；且同一数据组内不能有配置了**dynamicemb**的特征（dynamicemb复用了权重通道传递频次计数，不支持带权重的pooling）
 
 - **value_dim**: 默认值是0，可以设置1，value_dim=0时支持多值ID输出
 
--- a/docs/source/feature/feature.md
+++ b/docs/source/feature/feature.md
@@ -117,7 +117,7 @@ feature_configs {
 
 - **weighted**: 是否为带权重的Id特征，输入形式为`k1:v1\x1dk2:v2`
 
-- **use_weight**: 默认值是false，仅在**weighted**=true时生效。默认情况下，权重仅用于从`k:v`格式的输入中解析出id，解析后即被丢弃，不参与模型计算，也不会传输给模型。设置为true时，权重会作为per_sample_weights参与Embedding pooling；**pooling**=mean时，权重会按每个样本归一化到和为1，pooling以sum的方式完成，保证训练和推理结果一致。注意：开启后，同一**数据组（data group）**内的所有特征都会带上权重（其余特征填充为1）并走带权重的pooling kernel，会有一定的内存和计算开销；且同一数据组内不能有配置了**dynamicemb**的特征（dynamicemb复用了权重通道传递频次计数，不支持带权重的pooling）
+- **use_weight**: 默认值是false，仅在**weighted**=true时生效。默认情况下，权重仅用于从`k:v`格式的输入中解析出id，解析后即被丢弃，不参与模型计算，也不会传输给模型。设置为true时，权重会参与embedding pooling。注意：开启后，同一**数据组（data group）**内的所有特征都会带上权重（其余特征填充为1），会有一定的内存和计算开销；且同一数据组内不能有配置了**dynamicemb**的特征（dynamicemb目前不支持带权重的pooling）
 
 - **value_dim**: 默认值是0，可以设置1，value_dim=0时支持多值ID输出
 
--- a/tzrec/datasets/data_parser.py
+++ b/tzrec/datasets/data_parser.py
@@ -20,6 +20,7 @@ import pyarrow.compute as pc
 import pyfg
 import torch
 from torchrec import JaggedTensor, KeyedJaggedTensor, KeyedTensor
+from torchrec.modules.embedding_configs import PoolingType
 
 from tzrec.acc.utils import (
     is_cuda_export,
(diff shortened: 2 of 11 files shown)


tiankongdeguiji:
Thanks — I verified each finding against the code. Four are real and fixed in 8e68ba39; one is not, details below.

### Fixed

**1a — weighted sequence id feature.** Confirmed. Sequence sub-features are real `IdFeature` instances (`create_class(sub_feat_cls_name)(..., is_sequence=True)`, feature.py:1239-1244), so the flags get set, but the sequence branch never writes `{name}.weights` and `to_batch` died with a `KeyError` in a dataloader worker. `IdFeature.__init__` now raises. I gated on `weighted` rather than `use_weight` so the pre-existing trap is covered too — that cannot regress a working config, since any such config already crashed at the first batch.

**1c — stub features.** Confirmed, and it is worse than a mismatch: with a stub carrying `use_weight` and a mean-pooled feature in the same data group, that feature's table was rewritten to `SUM` and its collection built `is_weighted=True` against a KJT with no weights. `create_features` now applies the same `not x.stub_type` filter `DataParser.__init__` already uses.

**2 — docs.** Both correct. Now says 数据组（data group）to match what is enforced, the default-behaviour sentence no longer contradicts itself (the weight *is* split off the value to extract the id, then dropped), and it picks up your informational note about the group-wide cost.

**3 — test coverage.** All three gaps were real. Added: `INPUT_TILE=2` and `INPUT_TILE=3` tests asserting exact KJT weights (these set `INPUT_TILE_3_ONLINE=1` — without it `tile_size` is 1 and the tiling is a no-op, which is why the existing `test_input_tile*` would not have covered this even with a weighted feature); an all-zero-weight bag exercising the `weight_sum == 0` guard; and a per-collection `ebc_user`/`mc_ebc_user` test. Plus regression tests for the two guards above.

### Not a bug — 1b, dynamicemb on the sequence path

`ShardedDynamicEmbeddingCollection.input_dist` overwrites `features._weights` on **both** branches (`dynamicemb/shard/embedding.py:291-306`) — frequency counters in the `if`, `None` in the `else` — so weights tzrec put in the KJT are always discarded before the all-to-all. `_use_index_dedup` defaults to `False`, which takes the `else` branch, so "dedup disabled" is the safe case rather than a hole. `BatchedDynamicEmbedding.forward` does read `weights_or_none()`, but only from the post-`input_dist` KJT.

The guard is therefore correctly placed in `emb_bag_config`: the *pooled* `dynamicemb/shard/embeddingbag.py` has no `input_dist` override, which is exactly the corruption path it blocks. The residual cost of a dynamicemb sequence feature sharing a weighted data group is a wasted ones-tensor, not corrupted counters.

One correction on the citation: `multi_tower_din_fg_dynamicemb_mock.config` is not an example of dynamicemb-in-sequence. Across all test configs, every dynamicemb usage sits on a non-sequence top-level `id_feature` (`user_id`, `id_8`, `id_1`, `id_2`); that config's sequence block has none.

Verified on `torch 2.13.0 / torchrec 1.8.0` (matching the pins after the master merge): the four touched unit modules pass, plus `tzrec/features` (275) and `tzrec/datasets` (176) green, `pre-commit` and `pyrefly` clean.

eric-gecheng (approved):

chengaofei (approved):

PR #652, opened 2026-08-28
[feat] checkpoint success marker and list-column labels

tiankongdeguiji:
## Summary

Three gaps that show up when driving TorchEasyRec from an external scheduler, and from tables whose label columns are arrays.

- write a `_SUCCESS` marker into a checkpoint directory once the save completes, and record what that checkpoint is in its `meta`
- read labels through `BaseModel.get_label`, so a label stored as a list column is usable like an ordinary one
- document the determinism environment variables

## Motivation

**Checkpoint completion and identity.** A scheduler polling a model directory cannot tell a finished checkpoint from one still being written, so it either races the writer or guesses with a timeout. Having waited, it then needs to know *what the checkpoint is* — which step, and which data it covers. Neither was available: the only signals were the directory name, which several tools and user scripts parse and which is therefore not a safe place to add information, and the dataloader resume state, which is internal.

**List-column labels.** A label column of an array type (e.g. `array<bigint>` upstream) is parsed into a jagged label, and no model could consume one — every model reads `batch.labels[name]`, which does not contain it. The only way to train on such a column was to patch `DataParser` and flatten the label there.

### Completion and identity

A checkpoint now says what it is, and says when it is safe to read, using two files written in a deliberate order:

```
1. shards                  (collective, all ranks)
2. dataloader_state.json
3. meta                    <- step, data_ts, tag   (rank 0)
4. _SUCCESS                <- empty, last          (rank 0)
```

`meta` gains three descriptive keys beside the ones a converted checkpoint already carries:

```json
{"step": 1000, "data_ts": 1755000000.0, "tag": "20260828"}
```

- `step` — always present
- `data_ts` — the consumed event-time, **quorum-reconciled** so every rank records the same number. Omitted when the source has no event-time column, rather than a misleading sentinel. It is reconciled *after* the save decision, which is identical on every rank, so the gather stays in lockstep; it cannot affect the trigger, which was already decided. When the event-time triggers are on they have already reconciled this step, so the reporting path reuses that value: at most one gather per saved checkpoint, never one per step.
- `tag` — the optional `CHECKPOINT_TAG` environment variable, labelling which run produced a checkpoint. Omitted when unset.

`restore` is unaffected: it reads `meta` with `get("load_model", True)`, `get("load_optim", True)` and `get("dynamicemb_load_table_names", None)`, so the new keys are ignored and absent ones keep their defaults. A test asserts this.

`_SUCCESS` stays **empty** and is written **last**, by rank 0 behind a barrier — `CheckpointManager.save` runs on every rank and `save_model` is a collective `torch.distributed.checkpoint.save`, so the barrier is what stops the marker appearing before the last rank has finished its shards. Being empty and last is what makes it a reliable signal: its presence means every file beside it, `meta` and `dataloader_state.json` included, is complete. That removes any torn-read window without an atomic rename — which matters, because the fsspec monkeypatch does not cover `os.replace` and rename is not atomic on object stores, while `model_dir` is routinely `oss://` or `dfs://`. The ordering is load-bearing and the docstring says so.

The checkpoint directory name is deliberately left alone. Putting this information there was tried and abandoned (#653): the name is parsed by `_get_checkpoint_step`, `best_checkpoint`, pruning, several tools and users' own globs, and a `.` in a tag silently yields step 0 via `os.path.splitext`.

`__data_ts_watermark__` in `dataloader_state.json` is unchanged and still written only when an event-time trigger is configured. It serves a different purpose — continuing the save cadence across a restart, so a boundary or target in the gap is neither skipped nor repeated. Data resumption depends only on the per-source offsets and never used it.

### Labels

`BaseModel.get_label(batch, label_name)` returns `batch.labels[label_name]` when present and `batch.jagged_labels[label_name].values()` otherwise. Every model that reads a label for loss or metric now goes through it: `rank_model.py`, `multi_task_rank.py`, `rocket_launching.py` and `match_model.py`. `RocketLaunching` matters in particular — it is a `RankModel` subclass that overrides `loss` and `update_metric` with direct reads, so it would otherwise fail on a label its own base class handles.

Two direct reads are left deliberately: `dssm_v2.py` derives a batch size from the label's first dimension and `pepnet.py` reads a domain routing index. For a fixed-length list label, `values().size(0)` is `batch * length`, so routing those through the helper would replace today's `KeyError` with a quietly wrong number.

The lookup lives in `fx_util.fx_get_label` behind `@torch.fx.wrap`: a plain `label_name in batch.labels` breaks symbolic tracing, because `batch.labels` is a `Proxy` and `in` iterates it. That follows the existing pattern in `tzrec/modules/sequence.py`, which imports an `fx_*` helper and re-wraps it in the consuming module.

### Determinism

No code change was needed. Measured on torch 2.12.1+cu129: `torch.manual_seed` already seeds all CUDA devices, `cudnn.benchmark` already defaults to `False`, and with `cudnn.deterministic` left `False` the existing `USE_DETERMINISTIC_ALGORITHMS=1` alone gives bitwise-identical Conv1d backward over 30 runs. The environment variables are simply documented in the FAQ.

## Test Plan

- `tzrec/utils/checkpoint_util_test.py` — `meta` content for step, `data_ts` and `tag`, each omitted when absent, and leaving the restore-relevant keys at their defaults; the marker is empty; both skip on a non-zero rank; a step-triggered save records its event-time while leaving the resume watermark alone; a source without event-times records nothing
- `tzrec/datasets/data_parser_test.py` — a list-typed label column lands in `jagged_labels`, a scalar one in `labels` (this leg had no coverage before)
- `tzrec/models/rank_model_test.py`, `tzrec/models/match_model_test.py` — on both the rank and match paths, a jagged label drives loss and the metric to the same values as the equivalent dense label, in NORMAL and FX_TRACE modes
- `tzrec/tests/custom_model_integration_test.py` — the existing out-of-tree train_eval → eval → export run sets `CHECKPOINT_TAG` and, for every checkpoint, asserts the marker is zero bytes and that `meta` carries a `step` matching the directory plus the round-tripped tag
- `pre-commit run -a`, `pyrefly check` (0 errors)
- `tzrec/utils/checkpoint_util_test.py`, spawned two-rank tests counting `all_gather_object` — ranks consuming *different* event-times record the *same* one (the property that keeps a collective save in one directory), and the triggers-on path gathers once. The second fails `'None 2' != 'None 1'` without the reuse.
- full unit test suite on 1.4.0 (torch 2.13.0 / torchrec 1.8.0): **1673 tests, 0 failures**

## Upgrade note

`save_checkpoints_timestamp_quorum` is now validated whenever a save policy is set, not only when an event-time trigger is configured, because the reporting path uses the quorum as well. An explicit `save_checkpoints_timestamp_quorum: 0` therefore fails at startup where it was previously accepted and ignored. `docs/source/usage/train.md` already documents the valid range as `(0,1]`, so this enforces what was documented — worth a release note so it is not met at upgrade time.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -378,3 +378,16 @@ pip install -U "pyodps>=0.13.1"
 | `STORAGE_API_SDK_RETRY_TIMES`      | 5              | SDK内部重试次数（仅对连接错误生效）                                     |
 | `STORAGE_API_SESSION_POLL_TIMEOUT` | 600（秒）      | 等待read session离开INIT状态的超时时间                                  |
 | `STORAGE_API_COMPRESSION`          | `LZ4_FRAME`    | 传输压缩方式，可选`ZSTD`/`UNCOMPRESSED`                                 |
+
+**Q18: 如何复现实验**
+
+**解决方法：** TorchEasyRec通过环境变量控制随机性，不需要在模型或启动脚本里自己调用`torch.manual_seed`：
+
+```bash
+export PYTHONHASHSEED=0
+export CUBLAS_WORKSPACE_CONFIG=:4096:8
+export PYTHON_RANDOM_SEED=100007
+export NUMPY_MANUAL_SEED=100007
+export TORCH_MANUAL_SEED=100007          # 同时会设置所有CUDA设备的种子
+export USE_DETERMINISTIC_ALGORITHMS=1    # 已包含cudnn的确定性行为
+```
--- a/docs/source/feature/data.md
+++ b/docs/source/feature/data.md
@@ -320,6 +320,11 @@ pipeline.global-job-parameters: |
     label_fields: "buy"
   ```
 
+- label列的类型是数组时（例如上游表的类型为`array<bigint>`），会被解析成jagged label。
+  模型侧统一用`BaseModel.get_label`读取label，会自动取出其中的values，因此这类label
+  可以和普通label一样使用。自定义模型中读取label时也应使用`self.get_label(batch, label_name)`
+  而不是`batch.labels[label_name]`。
+
 ### sample_weight_fields
 
 - 训练样本的样本权重列名
--- a/docs/source/feature/data.md
+++ b/docs/source/feature/data.md
@@ -320,10 +320,8 @@ pipeline.global-job-parameters: |
     label_fields: "buy"
   ```
 
-- label列的类型是数组时（例如上游表的类型为`array<bigint>`），会被解析成jagged label。
-  模型侧统一用`BaseModel.get_label`读取label，会自动取出其中的values，因此这类label
-  可以和普通label一样使用。自定义模型中读取label时也应使用`self.get_label(batch, label_name)`
-  而不是`batch.labels[label_name]`。
+- label列的类型是数组时（例如上游表的类型为`array<bigint>`），会被解析成jagged label，
+  取其中的values后和普通label一样使用，无需额外配置
 
 ### sample_weight_fields
 
--- a/docs/source/models/user_define.md
+++ b/docs/source/models/user_define.md
@@ -142,6 +142,9 @@ python -m grpc_tools.protoc -I . -I "${TZREC_PATH}" \
 
 **模型模块必须在顶层 `import` 自己的 `*_pb2`。**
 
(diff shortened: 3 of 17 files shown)


tiankongdeguiji:
#653 is stacked on this branch and should merge after it.

eric-gecheng (dismissed):

eric-gecheng (approved):

PR #653, opened 2026-08-28
[feat] support a user-defined tag in the checkpoint name

tiankongdeguiji:
> **Stacked on #652 — merge that first.** This branch is cut from #652's head, so the diff below also
> contains its four commits. **Only the last two are this PR:**
> - `c56abe5d` `[bugfix] find the best checkpoint by step instead of rebuilding its name`
> - `82a7afed` `[feat] support a user-defined tag in the checkpoint name`
>
> Once #652 merges I will rebase onto master and the diff will shrink to those two.

## Summary

Let an external scheduler address a checkpoint by something other than the step.

- `CHECKPOINT_TAG` names checkpoints `model.ckpt-<tag>-<step>`
- the tag may contain `{data_ts}`, replaced by the event-time of the data the checkpoint covers
- fix `best_checkpoint`, which rebuilt a checkpoint's name from its step instead of looking it up

## Motivation

A scheduler often needs to find a checkpoint by the date of the data it covers, or by a run label. The
only ways today are to encode it in `--model_dir` or to rename the checkpoint directory — and renaming
breaks the step parsing that restore, pruning and best-export all depend on.

## Design

**The step stays last.** `_get_checkpoint_step` reads the last dash-separated token, so
`model.ckpt-<tag>-<step>` needs no parser change, sorting stays correct, and the tag may itself contain
dashes. Every consumer of the name was checked: the `latest_checkpoint` glob and sort, `_run_prune`,
and `_get_checkpoint_step` are all unaffected. Putting the tag *after* the step would break all of them.

**Configuration is an environment variable**, read in `CheckpointManager.__init__` via a new
`env_util.checkpoint_tag()` — so there is no proto field, no plumbing, and no change to `main.py`. It
also has to be rank-invariant, because the directory name is used by a collective write, and `torchrun`
propagates the environment to every rank.

**`{data_ts}` is taken from the quorum-reconciled watermark**, never the rank-local `data_timestamp`.
`maybe_save` already has the reconciled value in scope, so this is one optional argument rather than
new plumbing. Using the raw per-rank value would let ranks build divergent directory names on a
partitioned source, splitting a distributed checkpoint across directories. `{data_ts}` renders `0` on
sources with no event-time column.

**Tags that would break step parsing are rejected at startup.** A `.` in the tag silently yields step
`0`, because `_get_checkpoint_step` runs `os.path.splitext` first and a later dot stops `ext` matching
`.ckpt-`:

| tag | directory | parsed step |
|---|---|---|
| `20260828` | `model.ckpt-20260828-300` | 300 |
| `2026-08-28` | `model.ckpt-2026-08-28-300` | 300 |
| `v1.2` | `model.ckpt-v1.2-300` | **0** |

So the template is validated in `__init__` against `[A-Za-z0-9_-]` plus the literal `{data_ts}`. Since
that placeholder renders to digits, validating the template is sufficient and the failure is immediate
rather than at the first save. Loosening the parser instead was rejected: its `splitext` branch is
load-bearing for names carrying a trailing extension, and it is shared by restore, prune and export.

**`best_checkpoint`** read the best step from the eval-result file and then rebuilt
`model.ckpt-{step}`, which only works while every checkpoint is named exactly that. It now looks the
directory up by parsing each candidate's step — more robust even with no tag in play, and a
prerequisite for this feature.

## Test Plan

- `tzrec/utils/checkpoint_util_test.py` — naming with and without a tag; `{data_ts}` rendering and its
  `0` fallback; validation rejects `v1.2` / `1.4.0` / `a/b` / `a b` and accepts `20260828` /
  `2026-08-28` / `v1_2` / `d-{data_ts}`; `latest_checkpoint` orders a directory holding both tagged and
  untagged checkpoints; `_run_prune` keeps the newest N tagged plus the best; `best_checkpoint`
  resolves a tagged directory and errors when a metric step has no checkpoint. 72 tests pass.
- `tzrec/tests/custom_model_integration_test.py` — the existing two-rank train_eval → eval → export run
  now sets `CHECKPOINT_TAG` and asserts the tagged name, so this covers both that `torchrun` emits one
  directory (not one per rank) and that the restore path consumes it.
- `pre-commit run -a`, `pyrefly check` (0 errors)
- full unit test suite: running at the time of writing

## Caveat

The tag makes checkpoints **addressable, not sortable**. Ordering still comes from the step, which
restarts when a run warm-starts into a fresh `model_dir`, so "newest by step" remains meaningless
across date-partitioned directories. For "find the latest completed checkpoint", the `_SUCCESS` marker
plus mtime is the answer. Documented in `docs/source/usage/train.md`.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -378,3 +378,16 @@ pip install -U "pyodps>=0.13.1"
 | `STORAGE_API_SDK_RETRY_TIMES`      | 5              | SDK内部重试次数（仅对连接错误生效）                                     |
 | `STORAGE_API_SESSION_POLL_TIMEOUT` | 600（秒）      | 等待read session离开INIT状态的超时时间                                  |
 | `STORAGE_API_COMPRESSION`          | `LZ4_FRAME`    | 传输压缩方式，可选`ZSTD`/`UNCOMPRESSED`                                 |
+
+**Q18: 如何复现实验**
+
+**解决方法：** TorchEasyRec通过环境变量控制随机性，不需要在模型或启动脚本里自己调用`torch.manual_seed`：
+
+```bash
+export PYTHONHASHSEED=0
+export CUBLAS_WORKSPACE_CONFIG=:4096:8
+export PYTHON_RANDOM_SEED=100007
+export NUMPY_MANUAL_SEED=100007
+export TORCH_MANUAL_SEED=100007          # 同时会设置所有CUDA设备的种子
+export USE_DETERMINISTIC_ALGORITHMS=1    # 已包含cudnn的确定性行为
+```
--- a/docs/source/feature/data.md
+++ b/docs/source/feature/data.md
@@ -320,6 +320,11 @@ pipeline.global-job-parameters: |
     label_fields: "buy"
   ```
 
+- label列的类型是数组时（例如上游表的类型为`array<bigint>`），会被解析成jagged label。
+  模型侧统一用`BaseModel.get_label`读取label，会自动取出其中的values，因此这类label
+  可以和普通label一样使用。自定义模型中读取label时也应使用`self.get_label(batch, label_name)`
+  而不是`batch.labels[label_name]`。
+
 ### sample_weight_fields
 
 - 训练样本的样本权重列名
--- a/docs/source/models/user_define.md
+++ b/docs/source/models/user_define.md
@@ -142,6 +142,9 @@ python -m grpc_tools.protoc -I . -I "${TZREC_PATH}" \
 
 **模型模块必须在顶层 `import` 自己的 `*_pb2`。**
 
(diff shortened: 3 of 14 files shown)


tiankongdeguiji:
Closing in favour of recording this metadata in the `_SUCCESS` marker instead of the checkpoint directory name.

Encoding data in the directory name makes it a parsed contract: a `.` in a tag silently yields step 0 via `os.path.splitext` in `_get_checkpoint_step`, `best_checkpoint` had to change because it rebuilt the name from the step, and several tools and user scripts already assume `model.ckpt-<step>`. The `_SUCCESS` marker introduced in #652 is new, has no existing consumers, and is already the file a scheduler polls, so the metadata will go there and the directory name stays as it is.

PR #664, opened 2026-09-08
[WIP][bugfix] restore LR scheduler progress on resume

WhiteSwan1:
## Summary

Training checkpoints currently restore model and optimizer state but do not save LR scheduler progress, so resumed training restarts the learning rate schedule.

Add an opt-in `--restore_lr_scheduler` CLI flag, defaulting to false to preserve existing behavior. It works independently of `--ignore_restore_optimizer`. Checkpoints save per-rank scheduler state, and restoration also updates optimizer learning rates before the first resumed batch. Epoch-based schedules advance before the epoch checkpoint is saved.

For legacy checkpoints without scheduler state, reconstruct progress from the saved batch index or completed epoch count using the unchanged LR configuration. Report an error when the required progress is unavailable. Training with no existing checkpoint still starts normally.

The restore policy is a CLI option rather than a TrainConfig field: it is a launch-time decision, keeps existing configurations unchanged, and needs no protobuf change. Saved-state restoration requires the same world size, scheduler types/order, and local parameter-group counts.

## Test Plan

- Run scheduler, checkpoint, training-loop, optimizer, and prompt regression tests in the repository's PyTorch 2.13 environment: 179 tests passed. Coverage includes default behavior, cold start, independent optimizer restoration, legacy reconstruction, and step/epoch resume boundaries.
- Reuse the existing two-rank CPU/Gloo checkpoint test to verify rank-local learning rates and the next scheduler step after restoration.
- `pre-commit run -a`: all hooks passed. `pyrefly check`: 0 errors.

### DIFF
--- a/docs/source/usage/train.md
+++ b/docs/source/usage/train.md
@@ -20,6 +20,11 @@ torchrun --master_addr=localhost --master_port=32555 \
 - --fine_tune_checkpoint: 增量训练的checkpoint路径，如experiments/multi_tower_din_taobao_local/model.ckpt-0，如果不设置，增量训练使用model_dir下最近的检查点
 - --edit_config_json: 命令行以json的方式动态修改配置文件，如{"model_dir":"experiments/","feature_configs[0].raw_feature.boundaries":[4,5,6,7]}
 - --ignore_restore_optimizer: 加载checkpoint参数时，忽略加载优化器的参数
+- --restore_lr_scheduler: 从选中的checkpoint恢复学习率调度器，默认false，保持旧版本从头调度的行为。该参数独立于`--ignore_restore_optimizer`，后者仍只控制优化器状态的恢复。
+  - 续训时使用`--continue_train --restore_lr_scheduler`，恢复稀疏、稠密和part optimizer的调度进度及当前学习率；续训第一批即使用恢复后的学习率。新checkpoint会保存调度状态，恢复时要求world size、调度器类型/顺序及各rank参数组数量一致。
+  - 旧checkpoint没有调度状态时，按checkpoint步数（从0开始）或已完成epoch数重建，要求学习率配置与原训练一致。按epoch调度但缺少已完成epoch数等无法确定进度的情况会报错，不会静默从头调度。
+  - 不传此参数时，即使checkpoint中有调度状态，也不会加载；切换版本无需修改旧配置。按step调度仍以训练batch计数，不改变梯度累积下现有的调度口径。
+  - 此开关不写入`pipeline.config`，复现任务时需保留启动命令；恢复checkpoint时会在日志中记录优化器和调度器恢复开关。
 
 ### 环境变量
 
--- a/tzrec/main.py
+++ b/tzrec/main.py
@@ -389,6 +389,7 @@ def _train_and_evaluate(
     pipeline_config_path: Optional[str] = None,
     dense_ema: Optional[DenseEMA] = None,
     export_config: Optional[ExportConfig] = None,
+    restore_lr_scheduler: bool = False,
 ) -> None:
     """Train and evaluate the model."""
     is_rank_zero = int(os.environ.get("RANK", 0)) == 0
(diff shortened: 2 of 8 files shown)


WhiteSwan1:
Automated review of 2026-09-22 at head `1301e7d`. Branch head is now `17d5191`.
All ten inline comments and the summary sections are addressed below.

## Addressed

| Comment | Change |
|---|---|
| `main.py:579` | Scheduler restore gated on `restore_from_model_dir`. Fine-tune launches start the configured schedule and log a notice. `78142dd`; covered by `test_batch_resume[fine_tune]` |
| `checkpoint_util.py:1331` | Step-based reconstruction removed. It is not equivalent to the recorded position once a resume has run without the flag; a two-generation lineage measures the two 5× apart. `675d9f3` |
| `checkpoint_util.py:1306` | Rank-indexed layout replaced by one entry per scheduler; world-size check removed. Resharded and rescaled resume now succeed. `63534ce` |
| `checkpoint_util.py:1270` | Docstring rewritten; the collective it described was removed. `63534ce` |
| `main_test.py:565` | Class docstring added stating the replay invariant; comments added for the `4`, the `2×` and `pass_sizes`. Not yet committed |

## Not changed

| Comment | Reason |
|---|---|
| `checkpoint_util.py:1302` | 22 of 23 production `json.load` sites are unguarded, including `meta` (`:1079`) and `dataloader_state.json` (`:1437`), written by the same `save()`. A guard here alone would be inconsistent with the checkpoint path. Suitable as a separate file-wide change |
| `checkpoint_util.py:706` | Restore is position-only since `7d95f40`; no checkpoint configuration is applied, so the documented risk does not arise |
| `checkpoint_util.py:1326` | Two log lines per rank, once per job |

## No longer applicable

| Comment | Reason |
|---|---|
| `docs/source/usage/train.md:24` | The quoted sentence was removed; the flag's documentation is now one line. `lr_scheduler.json` remains absent from the checkpoint-contents section |
| `checkpoint_util_test.py:975` (a) | The named test was removed with the branch it guarded |
| `checkpoint_util_test.py:975` (b) | The stored payload is one integer for every scheduler class. `set_step(k)` is equivalent to *k* steps for all six subclasses at k = 0..17 |

## Notes for re-review

Two statements under *Verified independently* no longer describe the code: the
`all_gather_object` has been removed, and `step + 1` reconstruction holds only for
single-generation lineage. The off-by-one conventions, the byte-for-byte epoch-step result and
JSON-safety still hold; the payload is now one integer per scheduler.

Renamed since the review: `test_restore_progress` → `test_restore_position`,
`legacy_checkpoint` → `no_saved_state`. `_save_restore_worker` is unchanged in substance.

The branch conflicts with base `7479e85`. No check suites have run since `1301e7d`.

WhiteSwan1:
Thanks for the second pass at `72b7343`. The branch has moved on since — head is now `0b57518`, with upstream `master` at `d0e3ce5` merged in and the version bumped to 1.4.16 — so let me first say what changed, because it reshapes a few of the answers below.

## What changed after the review

The biggest change is how scheduler state is persisted (`2549983`). The review noted two docstring nits on `BaseLR.state_dict` (`lr_scheduler.py:40`, `:47`); while fixing them it became clear the override was doing less than it looked. Torch's own `state_dict()` just reads `__dict__` — harmless. The dangerous direction is `load_state_dict`, which is a bare `__dict__.update(...)`: it sets `last_epoch` but leaves `_step_count`, `_last_lr` and the optimizer's `lr` at the fresh scheduler's values, and every `_get_lr` in this file reads `_step_count`, not `last_epoch`. So `BaseLR` no longer overrides `state_dict` at all; `save_lr_schedulers` writes the one field a restore needs, and `BaseLR.load_state_dict` is torch's version followed by `set_step`, which realigns those three (`lr_scheduler.py:51-64`). Each entry in `lr_scheduler.json` is now `{"type": <class name>, "state": {"last_epoch": N}}` — the key is torch's own attribute name, and `type` sits outside the part handed to `__dict__.update` (`checkpoint_util.py:1220-1248`, `1251-1285`). Nothing had shipped with the old shape, so nothing migrates. The on-disk contract is documented in `save_lr_schedulers`' docstring now, beside the code that writes the file, which is where the `:47` comment wanted it anyway.

Smaller: the fine-tune notice is a warning and only fires when the flag is actually being ignored (`cbae919`); the unconditional info line is gone.

PR #669, opened 2026-09-14
[feat] support initial_accumulator_value on dynamicemb tables

tiankongdeguiji:
## Background

`initial_accumulator_value` (TensorFlow Adagrad's same-named parameter, which EasyRec users rely on) was already exposed on `FusedAdagradOptimizer`, but it only reached FBGEMM TBE tables. `SplitTableBatchedEmbeddingBagsCodegen` has no such kwarg and would raise on the `**fused_params` splat, so `create_sparse_optimizer` popped the key out of the optimizer kwargs and instead passed it through an env var consumed by the `apply_split_helper` patch.

dynamicemb tables read the very same fused params, so the key never reached them and `OptimizerArgs.initial_accumulator_value` stayed at its `0.0` default. A model with both kinds of tables therefore trained them with different Adagrad semantics, silently.

`FusedRowWiseAdagradOptimizer` had no such field at all, although both backends support one.

## Changes

- Carry the value to dynamicemb tables through `get_additional_fused_params()` of the customized-kernel parameter sharding. torchrec merges that per table *after* the optimizer kwargs, dynamicemb strips only its own planner keys, and `BatchedDynamicEmbeddingTables` takes `initial_accumulator_value` as a plain kwarg — so no dynamicemb-side change and no new monkeypatch.
- Add `initial_accumulator_value` to `FusedRowWiseAdagradOptimizer`. dynamicemb's `EXACT_ROWWISE_ADAGRAD` honours it, and the existing FBGEMM patch already keys on the `momentum1` state prefix, so rowwise adagrad gets it on both backends.
- Move the pop into the shared prologue of `create_sparse_optimizer` so both Adagrad variants are handled in one place, and wrap the env var in named accessors instead of repeating the FBGEMM string at the new call site.

### Alternatives considered

Letting the key stay in the optimizer kwargs and stripping it inside the FBGEMM kernel would have needed a second monkeypatch covering every `**fused_params` splat site (split TBE, SSD TBE, ...). Sharder-level `fused_params` were not usable either: `DynamicEmbeddingBagCollectionSharder` subclasses `EmbeddingBagCollectionSharder` and shards the whole EBC, so the value would have leaked onto the FBGEMM tables in the same collection. The per-table customized-kernel hook is the only place the two backends are actually distinguishable.

## Test Plan

- New `optimizer_builder_test` case: adagrad / rowwise adagrad / unset — the key never reaches the returned kwargs and the recorded value matches the config.
- New `plan_util_test.PlanUtilDynamicEmbE2ETest` case: the generated sharding plan's parameter sharding carries the value in its additional fused params, dynamicemb's `pop_additional_fused_params` leaves it in place, and `BatchedDynamicEmbeddingTables` still declares the kwarg (so an upstream rename fails loudly instead of silently dropping the value).
- Verified on GPU with a single-rank `DistributedModelParallel` shard that the real dynamicemb table's `OptimizerArgs.initial_accumulator_value` is the configured `0.1`, for both `adagrad_optimizer` and `rowwise_adagrad_optimizer`, and that a plain FBGEMM rowwise-adagrad table's `momentum1` buffer is filled with `0.1`.
- `multi_tower_din_fg_dynamicemb_mock.config` now sets `initial_accumulator_value: 0.1`, so `test_multi_tower_din_with_dynamicemb_train_eval` exercises the path end to end — passes.
- `test_multi_tower_din_with_fg_adagrad_init_acc_train_eval_export` (the FBGEMM regression) and all of `tzrec/optim`, `tzrec/utils/plan_util_test`, `tzrec/utils/dynamicemb_util_test`, `tzrec/main_test` pass.
- `pre-commit run` on the touched files and `pyrefly check` are clean.

https://claude.ai/code/session_013vLyBqTcb9nAcDbXN37nUs


tiankongdeguiji:
Addressed the review in bb959d7a.

**Fixed**

1. **`optimizer.py` — no direct coverage of the `momentum1` fill.** Real gap, and the one worth closing: this PR both reworked the gate and extended the mechanism to rowwise adagrad. Added `SparseInitAccumulatorValueTest` in `tzrec/optim/optimizer_test.py`, covering `EXACT_ADAGRAD` (accumulator per element) and `EXACT_ROWWISE_ADAGRAD` (per row) — it asserts the momentum1 buffer contents and width, that `weights` is left alone (which pins the `prefix == "momentum1"` gate), and that a `0.0` value leaves the buffer zeroed. Rather than the suggested fake `split` / callbacks, it builds a real CPU `SplitTableBatchedEmbeddingBagsCodegen`, so it also proves fbgemm still routes through the patched helper, and it needs no GPU lane. Mutation-checked: forcing the gate false, or flipping the prefix to `momentum2`, both fail it.

2. **`dynamicemb_util.py` — implicit "set before planning" contract.** Documented on `set_sparse_init_accumulator_value`, mirroring `set_auto_retain_evicted_keys`. Confirmed the impact note is right and benign: `apply_optimizer_in_backward` runs only in `train_and_evaluate` (main.py:889, before `create_planner` at :894); `evaluate` and `predict_checkpoint` build no in-backward optimizers, so their dynamicemb tables are `training=False` and hold no optimizer state for the default to affect.

3. **`optimizer.md` — both wording issues.** The note now reads `TF默认0.1，TorchEasyRec默认0.0`, and splits the semantics: a regular table fills the whole accumulator at table build time, a dynamicemb table fills a key's accumulator on first write.

4. **Nit: the 3-line inline comment.** Folded into the setter docstring, leaving one line at the call site. Also dropped the surrounding `if`, so a non-Adagrad optimizer resets the value instead of inheriting a stale one.

**Not changed**

5. **Nit: the `tzrec/version.py` bump.** Keeping it — the last five commits touching that file (18d185ee, 49501565, c3e9df85, 2a117eb7, b30998a8) are all ordinary feature/bugfix PRs that bump it in place, and `master` is at 1.4.7 so 1.4.8 is conflict-free.

Re-ran after the change: the new CPU test, `optimizer_builder_test`, `plan_util_test`, `dynamicemb_util_test`, `main_test`, and both GPU e2e runs (`test_multi_tower_din_with_dynamicemb_train_eval`, `test_multi_tower_din_with_fg_adagrad_init_acc_train_eval_export`) — all green; `pre-commit` and `pyrefly check` clean.

https://claude.ai/code/session_013vLyBqTcb9nAcDbXN37nUs

WhiteSwan1 (approved):

PR #681, opened 2026-09-20
[bugfix] restore ODPS sessions by partition index on resume

tiankongdeguiji:
## Summary

`OdpsReader._restore_sessions` rebuilt the per-input session list on resume as `restored_sess_reqs + current_sessions[n_restored:]`, which assumes the sessions found in the checkpoint are exactly the leading partitions in order. That does not hold:

- A partition that produced no rows before the checkpoint (e.g. an empty one under `is_orderby_partition`) never gets a checkpoint key, so the consumed set skips it.
- Session order in `dataloader_state.json` follows the rank-major merge in `save_dataloader_state`, not partition order.

With partitions `[P0, P1(empty), P2]` consumed into `P2`, restore produced `[s0, s2, s2_new]`: `P2`'s remaining rows were read via the old session and then `P2` was read again in full via the fresh session, with no error.

The `get_read_session` response carries no partition spec, so the reader cannot map a session back to its partition on its own. The checkpoint source key now carries the session position, `{input_path}#{sess_idx}#{session_id}:{start}`, and restore puts each validated session back at that index. Unconsumed partitions keep their freshly created session at their own position.

Error handling on resume:
- A key of one of the job's own input paths without a session index comes from an older version and raises a clear `RuntimeError` asking to restart from scratch (those sessions expire within a day of the writer dying anyway).
- Keys of unknown input paths keep the existing warn-and-skip, so a `fine_tune_checkpoint` state referencing other tables, or parquet paths containing `#`, still do not abort startup.
- An index beyond the current session list for the same input path (e.g. `is_orderby_partition` toggled off) raises.

Alternative considered: putting the partition spec in the key instead of the index. The index is simpler to parse and is what positional restore needs.

## Test Plan

- New `test_odps_dataset_checkpoint_resume_empty_partition` (ODPS-credentialed): table with `dt=20240319` (10000 rows), an empty `dt=20240318`, and `dt=20240320` (10000 rows) under `is_orderby_partition=True`. Consumes into the third partition, restores into a fresh dataset, asserts sessions 0 and 2 are the checkpoint sessions and session 1 is the dataset's own fresh one, then reads to exhaustion and asserts consumed plus remaining rows equal 20000. This fails on master with the duplicate read.
- New offline `OdpsRestoreSessionsTest` with a stub client: positional placement with reserved keys mixed in, unknown-path warn-and-skip, old-format key raise, out-of-range index raise. Runs in every CI lane.
- `test_odps_dataset_checkpoint_resume_orderby_partition` adjusted for the new key layout; `test_odps_dataset_checkpoint_metadata` now pins the three-segment key.
- `pre-commit run --files ...` and `pyrefly check` clean. ODPS tests run on the CPU CI for this PR.

https://claude.ai/code/session_01MgZZz8Baa365GMyVuRFvZb


tiankongdeguiji:
Addressed the four inline comments in 13fa120:

- Old-format keys now raise only when they belong to one of this job's input paths; keys of unknown paths (fine-tune checkpoints on other tables, parquet paths containing `#`) keep the warn-and-skip.
- Docstring Raises clause covers all three error cases and the unknown-path skip.
- Added offline `OdpsRestoreSessionsTest` (stub client): positional placement, unknown-path skip, old-format raise, out-of-range raise.
- `test_odps_dataset_checkpoint_metadata` pins the `{input_path}#{sess_idx}#{session_id}` layout.

The restored sessions not being registered with `_refresh_sessions_daemon` is pre-existing and left for a follow-up.

eric-gecheng (approved):

